PR #1, opened 2019-08-30
Initial implementation of easypost Go package.

richshaffer:


README.md
```diff
@@ -1 +1,166 @@
...
+		return
+	}
+}
+```
+
+Development
+-----------
+
+### Releasing
+   1. Update Version constant in version.go.
```
bkeyoumarsi: let's add a CHANGELOG file and add a step here to update it on each release.

thwarted:
Remove PII.  Use XXX-555-1212 as a local phone number.

richshaffer:
> Remove PII. Use XXX-555-1212 as a local phone number.

Hi Andrew. Can you please add some more verbosity in your comment? What is PII? Where to use 555-1212? Just in the README example, or also in the tests? A note on why that's important might be nice, too.

thwarted:
> What is PII? Where to use 555-1212? Just in the README example, or also in the tests? A note on why that's important might be nice, too.

PII is Personally Identifiable Information.  Some phone numbers, like 415-456-7890, look fake but probably aren't.  Sorry, I was unfamiliar with the names and addresses you've used, although there is apparently a mix of real (Lotta Hitschmanova, although dead) and fictional names (Danny Tanner).  And it's probably better to use real addresses (which may be necessary for things like address verification) of office buildings than of residences, or where addresses don't matter to use fully, obviously fictional addresses.

CHANGELOG
```diff
@@ -0,0 +1,3 @@
+=== 1.0.0 2019-09-03
+
+ * Initial release.
```
Roguelazer: all files should end with a newline

README.md
```diff
@@ -1 +1,167 @@
-This repository will hold [Go](https://golang.org/) bindings to the [EasyPost API](https://www.easypost.com/docs/api).
+# EasyPost Go Client Library
+
+EasyPost is the simple shipping API. You can sign up for an account at <https://easypost.com>.
+
```
Roguelazer: please mention the license somewhere in the readme

PR #6, opened 2019-10-31
add travis build

brntbeer:
This PR adds a `travis-ci.yml` file to automate running of travis CI for every pull request.

however! Travis will complain that there's no API keys set. So one this PR merges, someone from EasyPost will need to configure some environment variables in the travis profile as described here: https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,2 @@
+language: go
+script: go get github.com/EasyPost/easypost-go &&  go test ./tests/
--- a/tests/insurance_test.go
+++ b/tests/insurance_test.go
@@ -46,6 +46,8 @@ func TestInsuranceCreation(t *testing.T) {
 			Amount:       "101.00",
 		},
 	)
+	require.NoError(err)
+
 	assert.NotNil(insurance.ToAddress)
 	assert.NotNil(insurance.FromAddress)
 	assert.NotNil(insurance.Tracker)


richshaffer:
Hi Brent,

Thanks for this contribution. The current tests are not ideal, as they are using actual API calls. Even though most of them are using a test key, it would still be better if running the tests didn't rely on our API server.

Before we set up Travis CI, I'd like to see if we can mock the API calls with canned responses, possibly with an easy way to update them as the API endpoints are extended. Please give me some time to work on that. I just wanted to let you know what is going on, so you didn't think that we are ignoring this request.

Thanks,

-Richard

brntbeer:
@richshaffer thanks for your response! Happy you took time to do that. I’m new to Go, but if I get some time to look into how other projects do it or coworkers of mine do, should I’ll check back with you or just commit to this branch?

richshaffer:
> I’m new to Go, but if I get some time to look into how other projects do it or coworkers of mine do, should I’ll check back with you or just commit to this branch?

Either action is fine. Thanks!

richshaffer:
Hi Brent. Thanks again for this suggestion and contributing. We went ahead and implemented integration with Travis CI, in a way that doesn't require configure API keys.

Thanks again,

-Richard

PR #7, opened 2020-01-17
add shipment options

scrivy:
add carrier insurance amount, certified mail, and registered mail shipment options

### DIFF
--- a/shipment.go
+++ b/shipment.go
@@ -81,6 +81,8 @@ type ShipmentOptions struct {
 	BillThirdPartyPostalCode string     `json:"bill_third_party_postal_code,omitempty"`
 	ByDrone                  bool       `json:"by_drone,omitempty"`
 	CarbonNeutral            bool       `json:"carbon_neutral,omitempty"`
+	CarrierInsuranceAmount   float64    `json:"carrier_insurance_amount,omitempty"`
+	CertifiedMail            bool       `json:"certified_mail,omitempty"`
 	CODAmount                float64    `json:"cod_amount,omitempty,string"`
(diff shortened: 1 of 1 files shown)


shipment.go
```diff
@@ -110,6 +112,8 @@ type ShipmentOptions struct {
 	PrintCustom1Code         string     `json:"print_custom_1_code,omitempty"`
 	PrintCustom2Code         string     `json:"print_custom_2_code,omitempty"`
 	PrintCustom3Code         string     `json:"print_custom_3_code,omitempty"`
+	RegisteredMail           bool       `json:"registered_mail,omitempty"`
+	RegisteredMailAmount     *float64   `json:"registered_mail_amount,omitempty"`
```
richshaffer: Does this need to be a pointer? `registered_mail = true` and we don't send `registered_mail_amount`, is `0` not the default?

shipment.go
```diff
@@ -81,6 +81,8 @@ type ShipmentOptions struct {
 	BillThirdPartyPostalCode string     `json:"bill_third_party_postal_code,omitempty"`
 	ByDrone                  bool       `json:"by_drone,omitempty"`
 	CarbonNeutral            bool       `json:"carbon_neutral,omitempty"`
+	CarrierInsuranceAmount   float64    `json:"carrier_insurance_amount,omitempty"`
```
richshaffer: The other `float64` values have 'string' in the json struct tag. I'm pretty sure this is because the EasyPost API will return floats as strings. It's likely we should do it for these new fields as well.

PR #10, opened 2020-03-17
Added Travis CI and added test info to README

### LINKED ISSUE #2: Add travis-ci integration
https://docs.travis-ci.com/user/languages/go

Justintime50:
This closes #2 and adds some missing info to the README about how to run tests. Rates have also changed since last year so I tweaked a single unit test to allow those to pass again.

1. 2 variables will need to be added to the Travis CI backend: `TEST_API_KEY` and `PROD_API_KEY`. 
2. This repo will also need to be enabled via Travis CI if it isn't already.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: go
+
+go:
+- "1.10"
+- 1.11.x
+- 1.12.x
+- 1.13.x
+- 1.x
+- master
+
+scripts:
+    - cd tests
+    - EASYPOST_TEST_API_KEY=$TEST_API_KEY EASYPOST_PROD_API_KEY=$PROD_API_KEY go test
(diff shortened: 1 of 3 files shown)


Roguelazer:
I've enabled this in travis-ci, so if you push a new commit onto this branch it should run.

att14:
@Justintime50 you mentioned this may need some work to merge, what do we need to do to be confident in the change?

Justintime50:
I believe @richshaffer wanted to swap these to mock tests instead of hitting our API.

This PR will also obviate #6.

att14:
I'd vote that having the tests run on Travis is an improvement. If we want to mock out web interactions in the future thats fine with me.

What do we mean by mock tests? Something like VCR is fine because we can detect if the content of the request changes, but straight mocks of the API might be brittle.

Justintime50:
Yeah something along the lines of VCR is ideal, I also agree that having tests are better than not - other CL's do testing this way as well for now. They should all probably be moved to VCR eventually.

PR #12, opened 2020-03-18
Added examples for various objects

Justintime50:
The Go library didn't have any examples like many of our other libraries. This PR adds various basic examples for objects like addresses, parcels, shipments, scanforms, and batches. Additionally, the example on the README threw various type errors so that example has been corrected in this PR too.

### DIFF
--- a/examples/addresses/create_address.go
+++ b/examples/addresses/create_address.go
@@ -1,0 +1,41 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/EasyPost/easypost-go"
+	"os"
+)
+
+func main() {
+	apiKey := os.Getenv("EASYPOST_API_KEY")
+	if apiKey == "" {
+		fmt.Fprintln(os.Stderr, "missing API key")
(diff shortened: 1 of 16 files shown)


examples/addresses/create_address.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "error creating to address:", err)
+		os.Exit(1)
+		return
+	}
+
+	prettyJSON, err := json.MarshalIndent(toAddress, "", "    ")
+    if err != nil {
+		fmt.Fprintln(os.Stderr, "error creating JSON:", err)
+    }
```
att14, line 39: This formatting appears to be off in all the files.

```suggestion
	if err != nil {
		fmt.Fprintln(os.Stderr, "error creating JSON:", err)
	}
```

Justintime50: Is this only off on this particular if statement? Shows as indented with 4 spaces in VS Code but after re-indenting it I found it was lying.

PR #17, opened 2020-06-04
Added rate retrieval and example

Justintime50:
Moved the rate struct to its own file and added rate retrieval functionality along with an example as all other CLs have rate retrieval.


examples/rates/retrieve_rate.go
```diff
@@ -0,0 +1,32 @@
...
+		fmt.Fprintln(os.Stderr, "error retrieving rate:", err)
+		os.Exit(1)
+		return
+	}
+
+	prettyJSON, err := json.MarshalIndent(rate, "", "    ")
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "error creating JSON:", err)
+	}
+	fmt.Printf("%s\n", string(prettyJSON))
```
thwarted, line 31: You can use `fmt.Println(string(prettyJSON))` here for the same effect.

examples/rates/retrieve_rate.go
```diff
@@ -0,0 +1,32 @@
...
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "error retrieving rate:", err)
+		os.Exit(1)
+		return
+	}
+
+	prettyJSON, err := json.MarshalIndent(rate, "", "    ")
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "error creating JSON:", err)
+	}
```
thwarted, line 30: Things would appear to be in an error state at this point, but this just continues on, no exit or return.

Justintime50: Not sure I follow, I ran this and it printed the rate object in JSON.

thwarted: If it can't marshal the struct into formatted JSON, it emits the error, and then continues on to print the value of prettyJSON.  This isn't strictly a problem and is probably unlikely, since prettyJSON should be empty, but it is of questionable style to continue on in the event of an error like this: it makes it look like you missed something.

Justintime50: Oh I see, it's missing:

```go
os.Exit(1)
return
```

Added and pushed! Thanks.

Issue #11, opened 2020-03-18
Cannot create Scanforms from Batch IDs


Justintime50:
It appears that you cannot create a scanform from a batch ID and currently can only create a scanform from shipment IDs. The API Supports both, this library should allow for both as well.

eg code only allowing shipments:
```go
// CreateScanForm creates a scan form for the given Shipments.
//	c := easypost.New(MyEasyPostAPIKey)
//	out, err := c.CreateScanForm("shp_1", "shp_2")
func (c *Client) CreateScanForm(shipmentIDs ...string) (out *ScanForm, err error) {
	err = c.post(nil, "scan_forms", newscanFormRequest(shipmentIDs...), &out)
	return
}
```

LoganSimonsen:
It kinda seems like this function was intended to hit the proper endpoint, but it's a GET instead of a POST? https://github.com/EasyPost/easypost-go/blob/master/batch.go#L187

but the comment makes it sound like this was meant to do the thing we want?

> // GetBatchScanForms generates a scan form for the batch.

Justintime50:
#23 will correct this.

PR #23, opened 2020-09-19
Change 'GetBatchScanForms' to be 'CreateBatchScanForms'

LoganSimonsen:
This PR is intended to solve issue #11 raised by Justin Hammond.

Per EasyPost API Docs for creating a scan_form from a Batch object, the method should be a POST and not a GET.

docs link: https://www.easypost.com/docs/api#manifesting-scan-form

Rather than making a new method, I modified this existing method as I believe it was originally intended to do the thing we want to do.

`// GetBatchScanForms generates a scan form for the batch.`

I tested this method and indeed it failed as a GET request, as expected based on the API docs. Changing the method to a POST request resulted in successfully creating a scan_form using a batch id. 

renamed both methods `GetBatchScanForms` and `GetBatchScanFormsWithContext`
to be `CreateBatchScanForms` and `CreateBatchScanFormsWithContext` to follow established naming conventions.

### DIFF
--- a/examples/batches/scan_form_batch/scan_form_batch.go
+++ b/examples/batches/scan_form_batch/scan_form_batch.go
@@ -16,13 +16,13 @@ func main() {
     }
     client := easypost.New(apiKey)
 
-    // ScanForm Batch
+    // Create a ScanForm for a Batch
     batch, err := client.CreateBatchScanForms(
         "batch_12345",
         "string",
     )
     if err != nil {
-        fmt.Fprintln(os.Stderr, "error creating batch:", err)
+        fmt.Fprintln(os.Stderr, "error creating ScanForm for Batch:", err)
         os.Exit(1)
         return
     }
--- a/examples/batches/scan_form_batch/scan_form_batch.go
+++ b/examples/batches/scan_form_batch/scan_form_batch.go
@@ -3,9 +3,8 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"os"
-
 	"github.com/EasyPost/easypost-go"
+	"os"
 )
 
 func main() {
@@ -20,6 +19,7 @@ func main() {
 	// Create a ScanForm for a Batch
 	batch, err := client.CreateBatchScanForms(
 		"batch_12345",
+		"string",
 	)
 	if err != nil {
 		fmt.Fprintln(os.Stderr, "error creating ScanForm for Batch:", err)
--- a/examples/batches/scan_form_batch/scan_form_batch.go
+++ b/examples/batches/scan_form_batch/scan_form_batch.go
@@ -1,37 +1,37 @@
 package main
 
 import (
-    "encoding/json"
-    "fmt"
-    "github.com/EasyPost/easypost-go"
-    "os"
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/EasyPost/easypost-go"
 )
 
 func main() {
-    apiKey := os.Getenv("EASYPOST_API_KEY")
-    if apiKey == "" {
-        fmt.Fprintln(os.Stderr, "missing API key")
-        os.Exit(1)
-        return
-    }
-    client := easypost.New(apiKey)
+	apiKey := os.Getenv("EASYPOST_API_KEY")
+	if apiKey == "" {
+		fmt.Fprintln(os.Stderr, "missing API key")
+		os.Exit(1)
+		return
+	}
+	client := easypost.New(apiKey)
 
-    // Create a ScanForm for a Batch
-    batch, err := client.CreateBatchScanForms(
-        "batch_12345",
-        "string",
-    )
-    if err != nil {
-        fmt.Fprintln(os.Stderr, "error creating ScanForm for Batch:", err)
-        os.Exit(1)
-        return
-    }
+	// Create a ScanForm for a Batch
+	batch, err := client.CreateBatchScanForms(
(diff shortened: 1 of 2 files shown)


Justintime50 (changes requested):
@LoganSimonsen We'll need the build to pass before we can merge this in. The following errors occurred when tests were run on Go 1.10, 1.11, and 1.12:

```bash
5.93s$ go test
260# github.com/stretchr/testify/assert
261../../../stretchr/testify/assert/assertions.go:1703:5: undefined: errors.Is
262../../../stretchr/testify/assert/assertions.go:1726:6: undefined: errors.Is
263../../../stretchr/testify/assert/assertions.go:1749:5: undefined: errors.As
264../../../stretchr/testify/assert/assertions.go:1766:7: undefined: errors.Unwrap
265../../../stretchr/testify/assert/assertions.go:1770:7: undefined: errors.Unwrap
266FAIL	github.com/EasyPost/easypost-go/tests [build failed]
267The command "go test" exited with 2.
```

eg: https://travis-ci.com/github/EasyPost/easypost-go/builds/185363183

I restarted the build just to make sure there wasn't any race conditions and no luck. I haven't pinpointed why this error is being thrown, is it possible this package being used is no longer supported on these older versions? We aren't even testing newer versions such as 1.14 and 1.15.

Let's get these builds passing and we can once again look into merging this.

richshaffer:
> @LoganSimonsen We'll need the build to pass before we can merge this in. The following errors occurred when tests were run on Go 1.10, 1.11, and 1.12:
> 
> ```shell
> 5.93s$ go test
> 260# github.com/stretchr/testify/assert
> 261../../../stretchr/testify/assert/assertions.go:1703:5: undefined: errors.Is
> 262../../../stretchr/testify/assert/assertions.go:1726:6: undefined: errors.Is
> 263../../../stretchr/testify/assert/assertions.go:1749:5: undefined: errors.As
> 264../../../stretchr/testify/assert/assertions.go:1766:7: undefined: errors.Unwrap
> 265../../../stretchr/testify/assert/assertions.go:1770:7: undefined: errors.Unwrap
> 266FAIL	github.com/EasyPost/easypost-go/tests [build failed]
> 267The command "go test" exited with 2.
> ```
> 
Looks like maybe the testify package has been updated and now only supports newer versions of Go. I don't see anything in this diff that could be responsible for the errors.

Justintime50:
@LoganSimonsen if you changed the `.travis.yml` file to be the following, your tests should pass:

```yml
language: go

go:
- 1.13.x
- 1.14.x
- 1.15.x
- 1.x
- master

scripts:
    - cd tests
    - go test
```

I went and confirmed that the `testify` package itself is only testing against 1.13 and newer and is therefore not supported on the older versions.

Although this may get your tests passing here and make your PR eligible, the team will need to determine if we want to drop support for 1.10, 1.11, and 1.12 on testing or if we need to instead swap out our testing library for another.

examples/batches/scan_form_batch/scan_form_batch.go
```diff
@@ -0,0 +1,37 @@
...
+    }
+    client := easypost.New(apiKey)
+
+    // ScanForm Batch
+    batch, err := client.CreateBatchScanForms(
+        "batch_12345",
+        "string",
+    )
+    if err != nil {
+        fmt.Fprintln(os.Stderr, "error creating batch:", err)
```
Justintime50, line 25: This should be `error creating scanform`

examples/batches/scan_form_batch/scan_form_batch.go
```diff
@@ -0,0 +1,37 @@
...
+func main() {
+    apiKey := os.Getenv("EASYPOST_API_KEY")
+    if apiKey == "" {
+        fmt.Fprintln(os.Stderr, "missing API key")
+        os.Exit(1)
+        return
+    }
+    client := easypost.New(apiKey)
+
+    // ScanForm Batch
```
Justintime50, line 19: This could be more explicit, something like `Create a scanform for a batch`

examples/batches/scan_form_batch/scan_form_batch.go
```diff
@@ -0,0 +1,37 @@
...
+    batch, err := client.CreateBatchScanForms(
+        "batch_12345",
+        "string",
```
Justintime50, line 22: Let's delete this, if instead you wanted to show that this could be another batch, let's do that instead with something like `batch_456`, the above batch: `batch_12345` also could better follow other examples and truncate at the third digit.

Justintime50 (changes requested):

examples/batches/scan_form_batch/scan_form_batch.go
LoganSimonsen: how about `error creating ScanForm for Batch:` to avoid having the same error message as create_scanform.go line 25?

Justintime50: Works for me!

Roguelazer:
If you rebase on master (and remove your conflicted `.travis.yml`), this will be better.

richshaffer (approved):
Looks fine. I don't know if anyone is using this part of the API, but we might want to bump the API version since renaming the method makes it a breaking change.

Justintime50:
@richshaffer fair; however, the previous implementation of this method doesn't work at all so my assumption is that nobody has been using it (successfully).

PR #27, opened 2021-05-22
feat: adds smartrate functionality

Justintime50:
Adds the new `/smartrate` endpoint/method on the `Shipment` object.

**NOTE:** Unlike our other client libraries using the Smartrate functionality, this will not return the `result` key but instead just the direct list of smartrates.

Simply call `GetShipmentSmartrates` and provide the shipment ID and you'll be returned a list of smartrates like so:

```json
[
        {
            "carrier": "USPS",
            "carrier_account_id": "ca_123...",
            "created_at": "2021-05-04T21:15:59Z",
            "currency": "USD",
            "delivery_date": null,
            "delivery_date_guaranteed": true,
            "delivery_days": null,
            "est_delivery_days": null,
            "id": "rate_123...",
            "list_currency": null,
            "list_rate": null,
            "mode": "test",
            "object": "Rate",
            "rate": 0.01,
            "retail_currency": null,
            "retail_rate": null,
            "service": "First",
            "shipment_id": "shp_123",
            "time_in_transit": {
                "percentile_50": 2,
                "percentile_75": 3,
                "percentile_85": 4,
                "percentile_90": 4,
                "percentile_95": 4,
                "percentile_97": 4,
                "percentile_99": 4
            },
            "updated_at": "2021-05-04T21:15:59Z"
        },
    ]
```

A full code example would look like this:

```golang
smartrates, err := client.GetShipmentSmartrates("shp_123...")
if err != nil {
    fmt.Fprintln(os.Stderr, "error retrieving smartrates:", err)
    os.Exit(1)
    return
}

prettyJSON, err := json.MarshalIndent(smartrates, "", "    ")
if err != nil {
    fmt.Fprintln(os.Stderr, "error creating JSON:", err)
    os.Exit(1)
    return
}
fmt.Println(string(prettyJSON))
```

### DIFF
--- a/rate.go
+++ b/rate.go
@@ -7,30 +7,36 @@ import (
 
 // A Rate contains information on shipping cost and delivery time.
 type Rate struct {
-	ID                     string     `json:"id,omitempty"`
-	Object                 string     `json:"object,omitempty"`
-	Mode                   string     `json:"mode,omitempty"`
-	CreatedAt              *time.Time `json:"created_at,omitempty"`
-	UpdatedAt              *time.Time `json:"updated_at,omitempty"`
-	Service                string     `json:"service,omitempty"`
-	Carrier                string     `json:"carrier,omitempty"`
-	CarrierAccountID       string     `json:"carrier_account_id,omitempty"`
-	ShipmentID             string     `json:"shipment_id,omitempty"`
-	Rate                   string     `json:"rate,omitempty"`
-	Currency               string     `json:"currency,omitempty"`
-	RetailRate             string     `json:"retail_rate,omitempty"`
-	RetailCurrency         string     `json:"retail_currency,omitempty"`
-	ListRate               string     `json:"list_rate,omitempty"`
-	ListCurrency           string     `json:"list_currency,omitempty"`
-	DeliveryDays           int        `json:"delivery_days,omitempty"`
-	DeliveryDate           *time.Time `json:"delivery_date,omitempty"`
-	DeliveryDateGuaranteed bool       `json:"delivery_date_guaranteed,omitempty"`
-	EstDeliveryDays        int        `json:"est_delivery_dats,omitempty"`
+	ID                     string         `json:"id,omitempty"`
+	Object                 string         `json:"object,omitempty"`
+	Mode                   string         `json:"mode,omitempty"`
+	CreatedAt              *time.Time     `json:"created_at,omitempty"`
+	UpdatedAt              *time.Time     `json:"updated_at,omitempty"`
+	Service                string         `json:"service,omitempty"`
+	Carrier                string         `json:"carrier,omitempty"`
+	CarrierAccountID       string         `json:"carrier_account_id,omitempty"`
+	ShipmentID             string         `json:"shipment_id,omitempty"`
+	Rate                   string         `json:"rate,omitempty"`
+	Currency               string         `json:"currency,omitempty"`
+	RetailRate             string         `json:"retail_rate,omitempty"`
+	RetailCurrency         string         `json:"retail_currency,omitempty"`
+	ListRate               string         `json:"list_rate,omitempty"`
+	ListCurrency           string         `json:"list_currency,omitempty"`
+	DeliveryDays           int            `json:"delivery_days,omitempty"`
+	DeliveryDate           *time.Time     `json:"delivery_date,omitempty"`
+	DeliveryDateGuaranteed bool           `json:"delivery_date_guaranteed,omitempty"`
+	EstDeliveryDays        int            `json:"est_delivery_dats,omitempty"`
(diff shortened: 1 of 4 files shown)


rate.go
```diff
@@ -28,6 +28,11 @@ type Rate struct {
 	EstDeliveryDays        int        `json:"est_delivery_dats,omitempty"`
 }
 
+type Smartrate struct {
+	Rate
+	TimeInTransit map[string]interface{} `json:"time_in_transit,omitempty"`
+}
+
```
qeubar, line 35: Go was never really built for inheritance and this type of association usually leads to a lot of complications. For example one would assume that you can instantiate a `Smartrate` struct and initialize the `Rate` attributes as well but one would not be able to to (https://play.golang.org/p/mWvVQ9L2eus).

On the other hand, the use of `map[string]interface{}` to define a well documented json attribute goes against Go and this client library implementation. Technically the `time_in_transit` attribute is part of the `Rate` object according to our documentation: https://www.easypost.com/docs/api#smartrate

Here is what I recommend doing:
* Define `type TimeInTransit struct`
* Add `TimeInTransit` as an attribute under `type Rate struct`
* Debatable, define `type SmartRate Rate` to prettify the client library implementation

rate.go
```diff
@@ -40,3 +45,22 @@ func (c *Client) GetRateWithContext(ctx context.Context, rateID string) (out *Ra
 	err = c.get(ctx, "rates/"+rateID, &out)
 	return
 }
+
+type getShipmentSmartratesResponse struct {
+	Smartrate *[]*Smartrate `json:"result,omitempty"`
+}
+
```
qeubar, line 52: do this inside the `GetShipmentSmartRates` like all the other client funcs: https://github.com/EasyPost/easypost-go/blob/master/shipment.go#L180

rate.go
```diff
@@ -40,3 +45,22 @@ func (c *Client) GetRateWithContext(ctx context.Context, rateID string) (out *Ra
...
+// GetShipmentSmartrates fetches the available smartrates for a shipment.
+func (c *Client) GetShipmentSmartrates(shipmentID string) (out []*Smartrate, err error) {
```
qeubar, line 54: Since we are defining these on a shipment, they should be defined in `shipment.go` like the other shipment dependent funcs

qeubar (changes requested):

rate.go
Justintime50: We also have precedence to do this outside throughout the CL, is one better than the other? If so, what makes it more advantageous than the alternative? 

eg: https://github.com/EasyPost/easypost-go/blob/master/shipment.go#L234-L237

rate.go
Justintime50: Makes sense

rate.go
Justintime50: Ah nice.

rate.go
qeubar: heh there are just inconsistencies in this library.

now that I'm looking further, the `XXXWithContext` functions seems to be the reason for these inconsistencies and are not setup ideally.

Here's how are we should approach this going forward:
```
func (c *Client) GetShipmentSmartrates(shipmentID string) (out []*Rate, err error) {
	return c.GetShipmentSmartratesWithContext(nil, shipmentID)
}

func (c *Client) GetShipmentSmartratesWithContext(ctx context.Context, shipmentID string) (out []*Rate, err error) {
        res :=  struct {
	       Smartrates *[]*Rate `json:"result,omitempty"`
        }{Smartrates: &out}
	err = c.get(ctx, "shipments/"+shipmentID+"/smartrate", &res)
	return
}
```

qeubar: we can clean up the rest of the func implementations to follow the above convention in a followup

Justintime50: Aha, good to know. I've updated this and will file an issue to correct these.

shipment.go
```diff
@@ -290,6 +290,21 @@ func (c *Client) GetShipmentRatesWithContext(ctx context.Context, shipmentID str
 	return
 }
 
+// GetShipmentSmartrates fetches the available smartrates for a shipment.
+func (c *Client) GetShipmentSmartrates(shipmentID string) (out []*Rate, err error) {
```
qeubar, line 294: My spidy senses tell me that we should capitalize `SmartRates` to be fully CamelCased but I could be convinced otherwise.

Justintime50: I haven't in other libraries, in my mind, `Smartrate` is a single object/word/thought

qeubar: i don't have any strong feelings eitherway so i'll leave this up to your discretion

qeubar (approved):

qeubar:
thanks for adding this new functionality!

jontsai (approved):
LGTM

PR #38, opened 2021-11-30
chore: swaps the ISC license for the MIT license

Justintime50:
Swaps the `ISC` license for the `MIT` license as they are functionally the same, but we use the MIT license on all other client libraries. This will keep our licensing consistent across our open source client library projects.


Justintime50:
Closing due to not having all the signoffs needed to swap the license from each contributor.

PR #39, opened 2021-11-30
chore: adds new go versions to test against on CI

Justintime50:
Adds Go 1.16 and 1.17 to the matrix of go versions we test against.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request: ~
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -2,28 +2,28 @@ name: CI
...
+      - name: install dependencies
+        run: go get -u golang.org/x/lint/golint
+      - name: build
+        run: go build .
+      - name: go vet
+        run: go vet .
+      - name: go lint
+        run: golint .
+      - name: run tests
+        run: cd tests && go test -v
```
fiendish, line 29: I think your editor must autoformat on save? Formatting separately and putting format changes in their own commits makes reviewing easier.

fiendish (approved):

Justintime50:
@fiendish we want to be good stewards. It can only help, and we should be testing against every version we support or claim to.

ajnassar (approved):

jontsai (approved):

PR #40, opened 2021-12-02
chore: adds linting to CI

Justintime50:
Lints the project on CI with `golangci-lint` to enforce it for future contributions. This is important because it was found that `golint` is deprecated and wasn't even catching linting errors on CI (see #41).

This tool includes:
- golint (via Staticcheck)
- gofmt
- go vet
- various other tools for Golang formatting/linting

Linting the project found the following:
- Various unused variables
- A misspelling of `omitempty` for the `TaxIdentifiers` object which is important because without it, we'd be populating an empty TaxIdentifiers object on all rate objects unintentionally
- And various other code cleanups including switching the `nil` context to `context.Background()` to signify it's meant to be an empty context.

I also had to introduce a new `SmartRate` struct because the `rate` fields differ between the `Rate` and `SmartRate` structs, tests could not pass without this change.

### DIFF
--- a/tests/insurance_test.go
+++ b/tests/insurance_test.go
@@ -34,7 +34,7 @@ func (c *ClientTests) TestInsuranceCreation() {
 	)
 	require.NoError(err)
 
-	insurance, _ := client.CreateInsurance(
+	insurance, err := client.CreateInsurance(
 		&easypost.Insurance{
 			ToAddress:    to,
 			FromAddress:  from,
@@ -43,6 +43,7 @@ func (c *ClientTests) TestInsuranceCreation() {
 			Amount:       "101.00",
 		},
 	)
+	require.NoError(err)
 	assert.NotNil(insurance.ToAddress)
 	assert.NotNil(insurance.FromAddress)
 	assert.NotNil(insurance.Tracker)
--- a/tests/insurance_test.go
+++ b/tests/insurance_test.go
@@ -34,7 +34,7 @@ func (c *ClientTests) TestInsuranceCreation() {
 	)
 	require.NoError(err)
 
-	insurance, err := client.CreateInsurance(
+	insurance, _ := client.CreateInsurance(
 		&easypost.Insurance{
 			ToAddress:    to,
 			FromAddress:  from,
--- a/tests/pickup_test.go
+++ b/tests/pickup_test.go
@@ -111,7 +111,7 @@ func (c *ClientTests) TestPickupBatch() {
 	require.NoError(err)
 	require.NotEmpty(pickup.PickupRates)
 
-	pickup, err = client.BuyPickup(pickup.ID, pickup.PickupRates[0])
+	_, err = client.BuyPickup(pickup.ID, pickup.PickupRates[0])
 	require.NoError(err)
 }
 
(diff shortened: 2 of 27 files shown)


tests/insurance_test.go
```diff
@@ -34,7 +34,7 @@ func (c *ClientTests) TestInsuranceCreation() {
 	)
 	require.NoError(err)
 
-	insurance, err := client.CreateInsurance(
+	insurance, _ := client.CreateInsurance(
```
qeubar, line 37: the error should be checked in an assert below:
```
require.NoError(err)
```

tests/pickup_test.go
```diff
@@ -158,7 +158,7 @@ func (c *ClientTests) TestSinglePickup() {
 	minDatetime := noonOnNextMonday()
 	maxDatetime := minDatetime.AddDate(0, 0, 1)
 
-	pickup, err := client.CreatePickup(
+	pickup, _ := client.CreatePickup(
```
qeubar, line 161: same here, we should add a `NoError` check here as well.

tests/shipment_test.go
```diff
@@ -214,7 +214,7 @@ func (c *ClientTests) TestShipmentSmartrates() {
 	require.NoError(err)
 	require.NotEmpty(shipment.Rates)
 
-	smartrates, err := client.GetShipmentSmartrates(shipment.ID)
+	smartrates, _ := client.GetShipmentSmartrates(shipment.ID)
```
qeubar, line 217: `require.NoError(err)`

qeubar (changes requested):
thanks for fixing the linter here.

one question, how confident are we on continued support of this new linter?

Justintime50:
@qeubar fixed your feedback.

Great question! I feel confident it'll be around awhile. They have commits every couple days, last release was a month ago, I've used this linter for a year or so on personal projects and have enjoyed it. It's also used by large companies like Google, AWS, Netflix, etc: https://golangci-lint.run/product/trusted-by/

The build is failing because of a discrepancy in return types of some fields with `SmartRate` (nice, adding in the error checks caught it!) Will publish this for final review once that's corrected internally.

nwithan8 (approved):

jchen293 (approved):

tests/shipment_test.go
```diff
@@ -215,14 +215,15 @@ func (c *ClientTests) TestShipmentSmartrates() {
 	require.NotEmpty(shipment.Rates)
 
 	smartrates, err := client.GetShipmentSmartrates(shipment.ID)
+	require.NoError(err)
 	assert.Equal(shipment.Rates[0].ID, smartrates[0].ID)
-	assert.Equal(smartrates[0].TimeInTransit.Percentile50, 1)
```
qeubar, line 219: curious as to why we removed the exact value checks?

these tests are using VCR so these values should remain the same.

Justintime50: This is true, though the rate at which we've been re-recording cassettes (we've made various breaking changes across the libraries such as switching from form encoding to JSON encoding so we re-record the whole test suite) is annoying to constantly update these assertions because although VCR keeps the values the same, when removed, these values vary widely across runs and updating 7 entries is taxing. 

We're switching the other libraries to have a looser assertion here as a convenience to faster testing loops.

qeubar: thanks for the explanation.

qeubar (approved):

PR #43, opened 2022-01-26
Fix spelling error on json metadata

dlsteuer:
OriginLocation has a typo in the json field metadata which prevents this field from being populated on api calls.


Justintime50 (approved):
Fantastic find! Thanks for submitting this.

jchen293 (approved):

PR #44, opened 2022-01-31
add support for tracker create_list.

jchen293:
## What?
Fix the current **CreateTrackerList** in the tracker class.
## Why?
Currently, our tracker **CreateTrackerList** doesn't accept the correct parameters to allow users to create trackers in bulk. This PR fixes the current **CreateTrackerList** method by changing the parameter to a map of maps.
## How?
Change the parameter from **Tracker object struct** to a **map of maps**.
## Testing?
Add a unit test and use the VCR to record a cassette.


tracker.go
```diff
@@ -127,43 +126,19 @@ func (c *Client) CreateTrackerWithContext(ctx context.Context, opts *CreateTrack
...
-	for i, key := 0, 0; i < len(opts); i++ {
-		if m := opts[i].toMap(); m != nil {
-			trackers[strconv.Itoa(key)] = opts[i].toMap()
-			key++
-		}
-	}
-	req := map[string]interface{}{"trackers": trackers}
-	return c.post(nil, "trackers/create_list", req, nil)
+// CreateTrackerList asynchronously creates multiple trackers.
+// Input a map of maps that contains multiple trackings code
```
Justintime50, line 130:
```suggestion
// Input a map of maps that contains multiple tracking codes
```

tracker.go
```diff
@@ -127,43 +126,19 @@ func (c *Client) CreateTrackerWithContext(ctx context.Context, opts *CreateTrack
...
 }
 
-// CreateTrackerList asynchronously creates multiple trackers. Only
-// TrackingCode, Carrier and IsReturn parameters are supported.
-func (c *Client) CreateTrackerList(opts ...*CreateTrackerOptions) error {
-	// Make a nested map:
-	// {
-	//     "trackers": {
-	//         "0": { "tracking_code": "EZ1000000001", "carrier": "USPS" },
-	//         "1": { "tracking_code": "EZ1000000002", "carrier": "USPS" }
```
Justintime50, line 137: Please leave the example in the code here.

tests/tracker_test.go
```diff
@@ -1,9 +1,29 @@
...
+
+	trackingCodeParam := make(map[string]interface{})
+	trackingCodes := []string{"EZ1000000001", "EZ1000000001", "EZ1000000001"}
+
+	for i := range trackingCodes {
+		trackingCodeParam[strconv.Itoa(i)] = map[string]string{
+			"tracking_code": trackingCodes[i],
+			"carrier":       "USPS",
+		}
+	}
```
Justintime50, line 21: Although this seems to work, it actually needs to allow for a map of maps allowing both a `tracking_code` parameter and a `carrier` parameter. This looks like it's taking an array of `tracking_code` which isn't correct. It needs to be individual maps of both `tracking_code` and `carrier`.

Justintime50 (changes requested):

tests/tracker_test.go
Justintime50: The reason this is important is this is a bulk tracking endpoint. It won't always be just one carrier. Someone may want to create 100 trackers from 10 different carriers and each tracking code would need to allow its own carrier to be associated for that reason. This looks like it only allows for a single carrier for the whole batch.

Justintime50: Nevermind, I misread this.

tests/tracker_test.go
```diff
@@ -1,9 +1,29 @@
 package easypost_test
 
 import (
+	"strconv"
+
 	"github.com/EasyPost/easypost-go"
 )
 
+func (c *ClientTests) TestTrackerCreateList() {
```
Justintime50, line 9: Usually when adding something to an already existing file, you append to the end. This test could be at the end.

tracker.go
```diff
@@ -139,31 +138,18 @@ func (c *Client) CreateTrackerList(opts ...*CreateTrackerOptions) error {
...
-	for i, key := 0, 0; i < len(opts); i++ {
-		if m := opts[i].toMap(); m != nil {
-			trackers[strconv.Itoa(key)] = opts[i].toMap()
-			key++
-		}
-	}
-	req := map[string]interface{}{"trackers": trackers}
-	return c.post(nil, "trackers/create_list", req, nil)
+	req := map[string]interface{}{"trackers": param}
+	//this endpoint does not return a response so we return true here
```
Justintime50, line 142: Let's keep really clean comments by always capitalizing the start of the phrase and giving spaces between the start of `//` and the comment itself.

tracker.go
```diff
@@ -139,31 +138,18 @@ func (c *Client) CreateTrackerList(opts ...*CreateTrackerOptions) error {
...
-			// This key doesn't seem to matter to the API:
-			trackers[strconv.Itoa(key)] = opts[i].toMap()
-			key++
-		}
-	}
-	req := map[string]interface{}{"trackers": trackers}
-	return c.post(ctx, "trackers/create_list", req, nil)
+func (c *Client) CreateTrackerListWithContext(ctx context.Context, param map[string]interface{}) (bool, error) {
+	req := map[string]interface{}{"trackers": param}
+	//this endpoint does not return a response so we return true here
```
Justintime50, line 151: Same here

Justintime50 (approved):
Nice work! The usage seems consistent with our other libraries. Thanks for the unit test!

nwithan8 (approved):

PR #45, opened 2022-02-03
add retrieveMe() method in the User class.

jchen293:
Add **retrieveMe()** helper method in the User class, the users can call this method to grab their profile. E2E local test was done for this PR.

### DIFF
--- a/user.go
+++ b/user.go
@@ -101,15 +101,15 @@ func (c *Client) DeleteUserWithContext(ctx context.Context, userID string) error
 	return c.del(ctx, "users/"+userID)
 }
 
-// RetrieveMe retrieve the current user profile.
+// RetrieveMe retrieves the current user profile.
 func (c *Client) RetrieveMe() (out *User, err error) {
-	c.get(nil, "users/", &out)
+	c.get(nil, "users", &out)
 	return
 }
 
 // RetrieveMeWithContext performs the same operation as RetrieveMe, but allows
 // specifying a context that can interrupt the request.
 func (c *Client) RetrieveMeWithContext(ctx context.Context) (out *User, err error) {
-	c.get(ctx, "users/", &out)
+	c.get(ctx, "users", &out)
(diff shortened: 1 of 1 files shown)


user.go
```diff
@@ -100,3 +100,16 @@ func (c *Client) DeleteUser(userID string) error {
 func (c *Client) DeleteUserWithContext(ctx context.Context, userID string) error {
 	return c.del(ctx, "users/"+userID)
 }
+
+// RetrieveMe retrieve the current user profile.
```
Justintime50, line 104:
```suggestion
// RetrieveMe retrieves the current user.
```

user.go
```diff
@@ -100,3 +100,16 @@ func (c *Client) DeleteUser(userID string) error {
...
+func (c *Client) RetrieveMe() (out *User, err error) {
+	c.get(nil, "users/", &out)
```
Justintime50, line 106:
```suggestion
	c.get(nil, "users", &out)
```

user.go
```diff
@@ -100,3 +100,16 @@ func (c *Client) DeleteUser(userID string) error {
...
+	return
+}
+
+// RetrieveMeWithContext performs the same operation as RetrieveMe, but allows
+// specifying a context that can interrupt the request.
+func (c *Client) RetrieveMeWithContext(ctx context.Context) (out *User, err error) {
+	c.get(ctx, "users/", &out)
```
Justintime50, line 113:
```suggestion
	c.get(ctx, "users", &out)
```

nwithan8 (approved):

user.go
Justintime50: This wasn't completely resolved - these aren't really "profiles", but I won't block on that either.

Justintime50 (approved):

user.go
jchen293: My bad I will address this in the next PR.

PR #46, opened 2022-02-04
Update user brand

jchen293:
# What
Added support for updating the user brand in the Golang library, updated the changelog, and fixed a typo.

# Why
This is one of the missing core features.

# How
Added methods for updating the user brand and add a new file for a code example. Add a new class for Brand.

# Testing
E2E local testing, no unit test was written because there is my personal information in the cassette.

### DIFF
--- a/brand.go
+++ b/brand.go
@@ -1,5 +1,6 @@
 package easypost
 
+// Brand contains data for a user brand
 type Brand struct {
 	BackgroundColor string `json:"background_color,omitempty"`
 	Color           string `json:"color,omitempty"`
--- a/brand.go
+++ b/brand.go
@@ -1,0 +1,13 @@
+package easypost
+
+type Brand struct {
+	BackgroundColor string `json:"background_color,omitempty"`
+	Color           string `json:"color,omitempty"`
+	Logo            string `json:"logo,omitempty"`
+	LogoHref        string `json:"logo_href,omitempty"`
+	Ad              string `json:"ad,omitempty"`
+	AdHref          string `json:"ad_href,omitempty"`
+	Name            string `json:"name,omitempty"`
+	UserID          string `json:"user_id,omitempty"`
+	Theme           string `json:"theme,omitempty"`
+}
--- a/user.go
+++ b/user.go
@@ -125,7 +125,7 @@ func (c *Client) UpdateBrand(params map[string]interface{}, userID string) (out 
 	return
 }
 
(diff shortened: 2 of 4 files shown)


nwithan8 (approved):

brand.go
```diff
@@ -0,0 +1,13 @@
+package easypost
+
+type Brand struct {
```
Justintime50, line 3: Please add a comment above this line explaining what a brand is like all other structs have

Justintime50 (changes requested):
Code looks good, let's just add a struct comment and then we can merge.

user.go
```diff
@@ -113,3 +116,20 @@ func (c *Client) RetrieveMeWithContext(ctx context.Context) (out *User, err erro
 	c.get(ctx, "users", &out)
 	return
 }
+
+// UpdateBrand updates the user brand.
+func (c *Client) UpdateBrand(params map[string]interface{}, userID string) (out *Brand, err error) {
+	newParams := map[string]interface{}{"brand": params}
+	updateBrandUrl := fmt.Sprintf("users/%s/brand", userID)
```
Justintime50, line 123: Let's update `updateBrandUrl` to `updateBrandURL` to keep the linter happy.

Justintime50 (changes requested):

user.go
```diff
@@ -113,3 +116,20 @@ func (c *Client) RetrieveMeWithContext(ctx context.Context) (out *User, err erro
...
+
+// UpdateBrand updates the user brand.
+func (c *Client) UpdateBrand(params map[string]interface{}, userID string) (out *Brand, err error) {
+	newParams := map[string]interface{}{"brand": params}
+	updateBrandURL := fmt.Sprintf("users/%s/brand", userID)
+	c.put(nil, updateBrandURL, newParams, &out)
+	return
+}
+
+// UpdateBrand performs the same operation as UpdateBrand, but allows
```
Justintime50, line 128: Go docs expect the first word of the doc to be the function name, so here: `UpdateBrand` should become `UpdateBrandWithContext`.

Justintime50 (changes requested):

user.go
nwithan8: Wow, Godocs are really underwhelming for a modern language: https://go.dev/blog/godoc

Justintime50 (approved):

PR #47, opened 2022-02-04
Add timeout option, default value for HTTP requests

nwithan8:
Up until now, we have not provided a way for the user to set a custom timeout option for HTTP requests in our Go library.

This PR includes:
- A new `Timeout` property on the `Client` class, allowing the user to set a custom timeout for HTTP requests.
- `Timeout` defaults to 30 seconds if not set by user.

Due to using Go's built-in HTTP module, this timeout includes both the outgoing request and parsing the response.

Confirmed working by changing value and forcing timeouts during tests.

### DIFF
--- a/client.go
+++ b/client.go
@@ -25,7 +25,7 @@ func init() {
 	defaultUserAgent = fmt.Sprintf(
 		"EasyPost/v2 GoClient/%s Go/%s OS/%s",
 		Version, runtime.Version(), runtime.GOOS)
-	defaultTimeout = 60
+	defaultTimeout = 60000
 }
 
 // A Client provides an HTTP client for EasyPost API operations.
@@ -45,7 +45,7 @@ type Client struct {
 	// UserAgent is a User-Agent to be sent with API HTTP requests. If empty,
 	// a default will be used.
 	UserAgent string
-	// Timeout specifies the time limit (in seconds) for requests made by this Client. The
+	// Timeout specifies the time limit (in milliseconds) for requests made by this Client. The
 	// timeout includes connection time, any redirects, and reading the
 	// response body.
 	Timeout int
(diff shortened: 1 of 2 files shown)


Justintime50:
Is there a way to distinguish a connection vs request timeout?

client.go
```diff
@@ -64,6 +70,14 @@ func (c *Client) userAgent() string {
 	return defaultUserAgent
 }
 
+func (c *Client) timeout() time.Duration {
+	// return timeout in duration in seconds
+	if c.Timeout > 0 {
+		return time.Duration(c.Timeout) * time.Second
+	}
+	return time.Duration(defaultTimeout) * time.Second
```
Justintime50, line 78: Can we return once and assign the timeout to a variable here? if/else?

nwithan8: Just following the same syntax as the other methods (i.e. `userAgent()`)

nwithan8: Resolved in [5b9c631](https://github.com/EasyPost/easypost-go/pull/47/commits/5b9c63119f3019cfb551a7fe991a4a88d9d16586)

Justintime50 (approved):

tests/client_test.go
```diff
@@ -0,0 +1,23 @@
...
+		APIKey:  tempClient.APIKey,
+		Timeout: timeout1,
+	}
+	assert := c.Assert()
+	// test that property has been set correctly in the EasyPost.Client instance
+	assert.Equal(client.Timeout, timeout1)
+	// override the Timeout property after initialization
+	client.Timeout = timeout2
+	// test that property has been changed correctly
+	assert.Equal(client.Timeout, timeout2)
```
jchen293, line 22: I think these two **assert.Equal** parameters are in the wrong order. It should be **expected, actual.**

jchen293 (changes requested):

tests/client_test.go
nwithan8: Resolved in [0e54b0d](https://github.com/EasyPost/easypost-go/pull/47/commits/0e54b0da03341f2519e43e28a721b507eb295f53)

jchen293 (approved):

PR #48, opened 2022-02-04
Lowest rate

jchen293:
Allow users to get the lowest rate for a shipment, this is one of the missing core features that we haven't implemented in our Go library.
Add **lowestRate()** method in the shipment class, convert the user input to a map, and then loop through the rate to find the lowest rate for the shipment. Add multiple unit tests to ensure the code is working in all situations. A code example is also added since we don't have an API doc for our Go language yet.


tests/lowest_rate_test.go
```diff
@@ -0,0 +1,107 @@
+package easypost_test
+
+import (
+	"github.com/EasyPost/easypost-go"
+)
+
+func (c *ClientTests) TestLowestRate() {
+	client := c.TestClient()
+	assert, require := c.Assert(), c.Require()
+	to, err := client.CreateAddress(
```
Justintime50, line 10: Let's not create every object required here, you should be able to pass in raw data to the shipment create request instead of making an API call for every object individually.

Meaning, instead of creating two addresses and a parcel and customs, just create a map of each of their data and pass that in to the shipment create call. 1 API call instead of 6.

tests/lowest_rate_test.go
```diff
@@ -0,0 +1,107 @@
...
+			CustomsInfo: customsInfo,
+		},
+	)
+	require.NoError(err)
+	rate, err := client.LowestRate(shipment)
+
+	require.NoError(err)
+	assert.Equal("18.63", rate.Rate)
+	assert.Equal("UPSDAP", rate.Carrier)
+	assert.Equal("UPSStandard", rate.Service)
```
Justintime50, line 87: These are flakey assertions because if ever we need to re-record this cassette, most likely the rate or the carrier will change as rates are constantly in motion. Is there something else less flakey we could assert against?

tests/lowest_rate_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+	carriers := []string{"DHLExpress", "USPS"}
+	rateWithCarriersFilter, err := client.LowestRateWithCarrier(shipment, carriers)
```
Justintime50, line 90: These should all be separate tests. Each functionality should be tested individually and not grouped together. So it looks like we need 4 total tests here, one for each new `LowestRate` function.

shipment.go
```diff
@@ -351,3 +354,53 @@ func (c *Client) RerateShipmentWithContext(ctx context.Context, shipmentID strin
...
+		currentRate, _ := strconv.ParseFloat(out.Rate, 32)
+		newRate, _ := strconv.ParseFloat(rate.Rate, 32)
+
+		if (out == Rate{} || currentRate > newRate) && newRate > 0 {
+			out = *rate
+		}
+	}
+
+	if (out == Rate{}) {
+		return out, errors.New("No rates found.")
```
Justintime50, line 402: Check the GitHub Action errors here with linting. We should follow the feedback it gives on this string.

shipment.go
```diff
@@ -351,3 +354,53 @@ func (c *Client) RerateShipmentWithContext(ctx context.Context, shipmentID strin
...
+	}
+
+	if services != nil {
+		for _, service := range services {
+			servicesMap[strings.ToLower(service)] = true
+		}
+	}
+
+	for _, rate := range shipment.Rates {
+
```
Justintime50, line 387: Nit: No need for this extra line after the beginning of a for loop.

shipment.go
```diff
@@ -351,3 +354,53 @@ func (c *Client) RerateShipmentWithContext(ctx context.Context, shipmentID strin
...
+		if len(carriersMap) > 0 && !carriersMap[strings.ToLower(rate.Carrier)] ||
+			len(servicesMap) > 0 && !servicesMap[strings.ToLower(rate.Service)] {
+			continue
+		}
+
+		currentRate, _ := strconv.ParseFloat(out.Rate, 32)
+		newRate, _ := strconv.ParseFloat(rate.Rate, 32)
+
+		if (out == Rate{} || currentRate > newRate) && newRate > 0 {
```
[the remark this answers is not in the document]
Justintime50: Why do we need to convert it to a float here?

Justintime50 (changes requested):
Please also update the PR description now that it's out for review.

shipment.go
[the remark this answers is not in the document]
Justintime50: Ah, this is done for the comparison only, but we still return the original rate to the user.

shipment.go
```diff
@@ -351,3 +354,52 @@ func (c *Client) RerateShipmentWithContext(ctx context.Context, shipmentID strin
...
+			out = *rate
+		}
+	}
+
+	if (out == Rate{}) {
+		return out, errors.New("No rates found.")
```
Justintime50, line 401: Let's fix the linting error here (look on the file changed tab to see what needs fixing.)

examples/shipments/lowest_rate/lowest_rate.go
```diff
@@ -0,0 +1,135 @@
...
+	apiKey := os.Getenv("EASYPOST_API_KEY")
+	if apiKey == "" {
+		fmt.Fprintln(os.Stderr, "missing API key")
+		os.Exit(1)
+		return
+	}
+	client := easypost.New(apiKey)
+
+	// Create and verify addresses
+	toAddress, err := client.CreateAddress(
```
Justintime50, line 21: Like the unit test, let's make the example use raw data passed into the fields of the Shipment object instead of making a half-dozen API calls.

Justintime50 (changes requested):

shipment.go
Justintime50: This error still needs to be resolved.

tests/lowest_rate_test.go
```diff
@@ -0,0 +1,75 @@
...
+	client := c.TestClient()
+	assert, require := c.Assert(), c.Require()
+	shipment := *GenerateTestShipment()
+	rate, err := client.LowestRate(&shipment)
+
+	require.NoError(err)
+	assert.Equal("1.0", rate.Rate)
+	assert.Equal("USPS", rate.Carrier)
+	assert.Equal("ParcelSelect", rate.Service)
+
```
Justintime50, line 33: These extra lines here should be removed from these test functions.

tests/lowest_rate_test.go
```diff
@@ -0,0 +1,75 @@
...
+	assert, require := c.Assert(), c.Require()
+	shipment := *GenerateTestShipment()
+	carrier := []string{"FedEx", "UPS"}
+	rate, err := client.LowestRateWithCarrier(&shipment, carrier)
+
+	require.NoError(err)
+	assert.Equal("2.0", rate.Rate)
+	assert.Equal("UPS", rate.Carrier)
+	assert.Equal("ParcelSelect", rate.Service)
+
```
Justintime50, line 47: These extra lines here should be removed from these test functions.

tests/lowest_rate_test.go
```diff
@@ -0,0 +1,75 @@
...
+func (c *ClientTests) TestNoRateAvailable() {
+	client := c.TestClient()
+	assert, require := c.Assert(), c.Require()
+	shipment := *GenerateTestShipment()
+	carrier := []string{"Cainiao"}
+	rate, err := client.LowestRateWithCarrier(&shipment, carrier)
+
+	require.Error(err)
+	assert.Equal(easypost.Rate{}, rate)
+
```
Justintime50, line 74: These extra lines here should be removed from these test functions.

tests/lowest_rate_test.go
```diff
@@ -0,0 +1,75 @@
+package easypost_test
+
+import (
+	"github.com/EasyPost/easypost-go"
+)
+
+func GenerateTestShipment() *easypost.Shipment {
+
```
Justintime50, line 8: Don't open a function with a line-break, it should immediately jump into the code.

tests/lowest_rate_test.go
```diff
@@ -0,0 +1,75 @@
+package easypost_test
```
Justintime50, line 1: All of the tests in this file should instead go in the `shipment_test.go` file since they are shipment tests. We try to group all tests into a file named after the object being tested. Each object may then have a dozen or more individual tests to test different functionality related to that object.

I'm also not seeing a cassette for these tests. Is VCR working properly to record these so that we can have consistent results across test runs? AH - after looking I see you're mocking the `Rates` object, pretty clever! With that being said, we should be making a real HTTP call to create a shipment and then getting the lowest rate from that instead so that we can test this completely.

Justintime50 (changes requested):

Justintime50 (approved):

tests/lowest_rate_test.go
jchen293: As per our conversation, we are going to mock the `Rates` object for now.

nwithan8 (approved):

PR #55, opened 2022-03-01
Add missing refund functions 

jchen293:
This PR adds missing functions in Refund.

- Add `CreateRefund`, `CreateRefundWithContext`, `ListRefunds`, `ListRefundsWithContext`, `GetRefund`, and `GetRefundWithContext` in the Refund file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,10 @@
 
 ## Next Release
 
-<<<<<<< HEAD
 * Add comprehensive test coverage
 * Fix the bug for `AddShipmentsToBatch`, `AddShipmentsToBatchWithContext`, change the parameter from `string` to `interface{}` so user can pass in either Shipment IDs or Shipment structs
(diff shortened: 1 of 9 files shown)


batch.go
```diff
@@ -77,7 +77,7 @@ func (c *Client) CreateAndBuyBatchWithContext(ctx context.Context, in ...*Shipme
 
 // ListBatchesResult holds the results from the list insurances API.
 type ListBatchesResult struct {
-	Insurances []*Insurance `json:"batches,omitempty"`
+	Batches []*Batch `json:"batches,omitempty"`
```
Justintime50, line 80: Please break the batch changes into a separate PR. PRs should be as small as possible and only contain changes related to one thing. This PR is for the `refund` object and should only have changes related to that.

Justintime50 (changes requested):
See last comment

Justintime50 (approved):

nwithan8 (approved):

PR #56, opened 2022-03-03
fix typo in Batch and add ID attribute for Brand struct.

jchen293:
Add `ID` attribute to the `Brand` struct. 
Fix a typo in the `Batch` file, the type of `ListBatchesResult` should be `batch` instead of `insurance`.

### DIFF
--- a/brand.go
+++ b/brand.go
@@ -2,6 +2,7 @@ package easypost
 
 // Brand contains data for a user brand
 type Brand struct {
+	ID              string `json:"id,omitempty"`
 	BackgroundColor string `json:"background_color,omitempty"`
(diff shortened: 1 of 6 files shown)


brand.go
```diff
@@ -11,4 +11,5 @@ type Brand struct {
 	Name            string `json:"name,omitempty"`
 	UserID          string `json:"user_id,omitempty"`
 	Theme           string `json:"theme,omitempty"`
+	ID              string `json:"id,omitempty"`
```
Justintime50, line 14: ID should be the first thing in the struct.

nwithan8 (approved):

PR #57, opened 2022-03-04
add create_and_verify endpoint functions in Address.

jchen293:
Add missing `CreateAndVerifyAddress` and `CreateAndVerifyAddressWithContext` functions in Address file
Add missing `message` attribute in Address struct
Add missing `Verifications` in `IncorrectAddressToVerify()` reusable function

### DIFF
--- a/address.go
+++ b/address.go
@@ -75,6 +75,10 @@ type createAddressRequest struct {
 	Address *Address `json:"address,omitempty"`
 }
 
+type AddressVerifyResponse struct {
+	Address *Address `json:"address,omitempty"`
+}
+
 // CreateAddress submits a request to create a new address, and returns the
 // result.
 //	c := easypost.New(MyEasyPostAPIKey)
@@ -89,7 +93,7 @@ type createAddressRequest struct {
 //			Company: "EasyPost",
 //			Phone:   "415-123-4567",
 //		},
-//		&CreateAddrssOptions{Verify: []string{"delivery"}},
+//		&CreateAddressOptions{Verify: []string{"delivery"}},
 //	)
 func (c *Client) CreateAddress(in *Address, opts *CreateAddressOptions) (out *Address, err error) {
 	req := &createAddressRequest{CreateAddressOptions: opts, Address: in}
@@ -159,5 +163,24 @@ func (c *Client) GetAddress(addressID string) (out *Address, err error) {
 // specifying a context that can interrupt the request.
 func (c *Client) GetAddressWithContext(ctx context.Context, addressID string) (out *Address, err error) {
 	err = c.get(ctx, "addresses/"+addressID, &out)
+	return
+}
+
+// CreateAndVerifyAddress Create Address object and immediately verify it.
+func (c *Client) CreateAndVerifyAddress(in *Address, opts *CreateAddressOptions) (out *Address, err error) {
+	req := &createAddressRequest{CreateAddressOptions: opts, Address: in}
+	response := AddressVerifyResponse{}
+	err = c.post(context.Background(), "addresses/create_and_verify", req, &response)
+	out = response.Address
+	return
+}
+
+// CreateAndVerifyAddressWithContext performs the same operation as CreateAndVerifyAddress, but allows
+// specifying a context that can interrupt the request.
+func (c *Client) CreateAndVerifyAddressWithContext(ctx context.Context, in *Address, opts *CreateAddressOptions) (out *Address, err error) {
+	req := &createAddressRequest{CreateAddressOptions: opts, Address: in}
+	response := AddressVerifyResponse{}
+	err = c.post(ctx, "addresses/create_and_verify", req, &response)
+	out = response.Address
 	return
 }
--- a/tests/address_test.go
+++ b/tests/address_test.go
@@ -86,6 +86,22 @@ func (c *ClientTests) TestAddressCreateVerify() {
 	assert.Equal("417 MONTGOMERY ST STE 500", address.Street1)
 }
 
+func (c *ClientTests) TestAddressCreateAndVerify() {
+	client := c.TestClient()
+	assert := c.Assert()
+
+	address, _ := client.CreateAndVerifyAddress(
+		c.fixture.IncorrectAddressToVerify(),
+		&easypost.CreateAddressOptions{
+			Verify: []string{"delivery"},
+		},
+	)
+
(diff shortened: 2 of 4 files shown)


tests/fixture_test.go
```diff
@@ -64,6 +64,11 @@ func (fixture *Fixture) IncorrectAddressToVerify() *easypost.Address {
 		Country: "US",
 		Company: "EasyPost",
 		Phone:   "415-123-4567",
+		Verifications: &easypost.AddressVerifications{
+			Delivery: &easypost.AddressVerification{
+				Success: true,
```
Justintime50, line 69: Why are we setting `Success` to `true` on the request data?

address.go
```diff
@@ -55,6 +55,7 @@ type Address struct {
 	Name            string                `json:"name,omitempty"`
 	Company         string                `json:"company,omitempty"`
 	Phone           string                `json:"phone,omitempty"`
+	Message         string                `json:"message,omitempty"`
```
Justintime50, line 58: Why is `message` being added here? Addresses don't have a message property.

address.go
```diff
@@ -75,6 +76,11 @@ type createAddressRequest struct {
 	Address *Address `json:"address,omitempty"`
 }
 
+type AddressVerifyResponse struct {
```
Justintime50, line 79: Please double check the documentation surrounding address verification. I don't believe this is the correct way to set this up. If you look at the recorded cassette, you'll notice that `message` does not appear here at all.

Justintime50 (changes requested):

tests/fixture_test.go
jchen293: this is for **IncorrectAddressToVerify** request data, I forgot to add this on the test suite PR.

address.go
jchen293: I was using Java for reference: https://github.com/EasyPost/easypost-java/blob/master/src/main/java/com/easypost/model/Address.java#L144-L157

The API is going to return a `message` and `Address`, we set the `message` from API call to `Address` if it's not null. Therefore, we need to create another struct that has two attributes `message` and `Address`.

address.go
jchen293: Address needs `message` attribute for `create_any_verify` endpoint, C# and Java have `message` attribute in their Address class.

C#: https://github.com/EasyPost/easypost-csharp/blob/master/EasyPost.Net/Address.cs#L28
Java: https://github.com/EasyPost/easypost-java/blob/master/src/main/java/com/easypost/model/Address.java#L24

address.go
Justintime50: Can you show me where the API returned a message and address? I don't see that in the recorded cassette.

address.go
Justintime50: Does it "need" it, or were you just following the C# and Java libs? You'll notice that the response does not contain `message` at all. I believe the `message` key is no longer valid/used. If you check our internal code for the address controller on the API level, there is no mention of a `message` key at all. I assume we previously exposed a bare `message` key, but have since wrapped it up in the `verifications` dictionary.

tests/fixture_test.go
Justintime50: You are altering `response` details here, not `request` details. The proper way to pass this parameter is by adding `&CreateAddressOptions{Verify: []string{"delivery"}},` like the `CreateAddress` example shows in the inline comment right above it. 

It also appears there is a missing `e` in `Address` in that example, please also correct that.

address.go
jchen293: Yes, but we are not going to use the `AddressVerificationFieldError` struct because it's for something else and does not have the `Address` attribute.

tests/fixture_test.go
jchen293: You are right, `CreateAndVerifyAddress` is missing a second parameter for verification.

address.go
jchen293: After some internal discussion, the `message` attribute is no longer needed, this also applies to other libraries and we will address them in the future.

address.go
jchen293: After internal discussion, we found out that this endpoint returns the verified address in a nested dictionary, we don't need the `message` attribute in this struct.

tests/address_test.go
```diff
@@ -86,6 +86,22 @@ func (c *ClientTests) TestAddressCreateVerify() {
 	assert.Equal("417 MONTGOMERY ST STE 500", address.Street1)
 }
 
+func (c *ClientTests) TestAddressCreateAndVerify() {
+	client := c.TestClient()
+	assert := c.Assert()
+
+	address, _ := client.CreateAndVerifyAddress(
+		c.fixture.IncorrectAddressToVerify(),
+		&easypost.CreateAddressOptions{
```
Justintime50, line 95: This should probably be on the fixture level as it is in other libs, you may need to remove other tests that have the option in the test itself and just move it to the fixture: https://github.com/EasyPost/easypost-php/blob/b654b3e9b43ef37992727fb48b8799df99dd5a98/test/EasyPost/Fixture.php#L63

address.go
```diff
@@ -161,3 +165,28 @@ func (c *Client) GetAddressWithContext(ctx context.Context, addressID string) (o
...
+
+	out = response.Address
+	return
+}
+
+// CreateAndVerifyAddressWithContext performs the same operation as CreateAndVerifyAddress, but allows
+// specifying a context that can interrupt the request.
+func (c *Client) CreateAndVerifyAddressWithContext(ctx context.Context, in *Address, opts *CreateAddressOptions) (out *Address, err error) {
+	req := &createAddressRequest{CreateAddressOptions: opts, Address: in}
+
```
Justintime50, line 185: Let's not do line breaks between each of these lines just so we match the style of the rest of this repo.

Justintime50 (changes requested):
Nice work! Couple tiny style things but the logic looks good. Test actually returns what it's supposed to.

tests/address_test.go
jchen293: This would roll back to the previous commit code, is this fine?

`
Verifications: &easypost.AddressVerifications{
			Delivery: &easypost.AddressVerification{
				Success: true,
       }
}
`

nwithan8 (approved):

tests/address_test.go
Justintime50: After talking, I misunderstood. This is correct and because the options is a parameter, we can't include it in the fixture.

Justintime50 (approved):

PR #61, opened 2022-03-09
Fix batch create and retrieve test param.

jchen293:
Fixes a couple of batch test params.


Justintime50:
This isn't necessary as the fixture being used is already correct.

PR #63, opened 2022-03-22
[feat] Support for `columns` and `additional_columns` on Report creation

nwithan8:
This PR:
- Adds support for specifying `Columns` and `AdditionalColumns` (both lists of strings) in parameters for the `CreateReport` function.

Unfortunately, since the API response does not indicate the columns used for a report, the only way to ensure the report actually includes the columns requested would be to make a live API call, download the generated CSV report and parse its column headers. That is a complex unit test with many points of failure and prevents us from using VCR.

Instead, we simply assume that if no error is thrown during the API call to generate and retrieve the report, that the report contains the correct columns. If it doesn't, that constitutes a server-side error that is out of our control.

This has been end-to-end tested locally. The generated report included the requests columns.

### DIFF
--- a/report.go
+++ b/report.go
@@ -9,23 +9,25 @@ import (
 // Report represents a CSV-formatted file that is a log of all the objects
 // created within a certain time frame.
 type Report struct {
-	ID              string     `json:"id,omitempty"`
-	Object          string     `json:"object,omitempty"`
-	Mode            string     `json:"mode,omitempty"`
-	CreatedAt       *time.Time `json:"created_at,omitempty"`
-	UpdatedAt       *time.Time `json:"updated_at,omitempty"`
-	Status          string     `json:"status,omitempty"`
-	StartDate       string     `json:"start_date,omitempty"`
-	EndDate         string     `json:"end_date,omitempty"`
-	IncludeChildren bool       `json:"include_children,omitempty"`
-	URL             string     `json:"url,omitempty"`
-	URLExpiresAt    *time.Time `json:"url_expires_at,omitempty"`
-	SendEmail       bool       `json:"send_email,omitempty"`
+	ID                string     `json:"id,omitempty"`
+	Object            string     `json:"object,omitempty"`
+	Mode              string     `json:"mode,omitempty"`
+	CreatedAt         *time.Time `json:"created_at,omitempty"`
+	UpdatedAt         *time.Time `json:"updated_at,omitempty"`
+	Status            string     `json:"status,omitempty"`
+	StartDate         string     `json:"start_date,omitempty"`
+	EndDate           string     `json:"end_date,omitempty"`
+	IncludeChildren   bool       `json:"include_children,omitempty"`
+	URL               string     `json:"url,omitempty"`
+	URLExpiresAt      *time.Time `json:"url_expires_at,omitempty"`
+	SendEmail         bool       `json:"send_email,omitempty"`
+	Columns           []string   `json:"columns,omitempty"`
+	AdditionalColumns []string   `json:"additional_columns,omitempty"`
 }
 
 // CreateReport generates a new report. Valid Fields for input are StartDate,
 // EndDate and SendEmail. A new Report object is returned. Once the Status is
-// available, the report can be downloded from the provided URL for 30 seconds.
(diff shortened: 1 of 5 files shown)


report.go
```diff
@@ -23,23 +23,36 @@ type Report struct {
...
 //	c := easypost.New(MyEasyPostAPIKey)
 //	c.CreateReport(
 //		"payment_log",
 //		&easypost.Report{StartDate: "2016-10-01", EndDate: "2016-10-31"},
 //	)
-func (c *Client) CreateReport(typ string, in *Report) (out *Report, err error) {
-	err = c.post(context.Background(), "reports/"+typ, in, &out)
+func (c *Client) CreateReport(typ string, in *Report, opts *CreateReportOptions) (out *Report, err error) {
+	req := &createReportRequest{CreateReportOptions: opts, Report: in}
+	err = c.post(context.Background(), "reports/"+typ, req, &out)
```
Justintime50, line 47: `typ` should be `type`, I know you didn't do it but let's fix it.

report.go
```diff
@@ -23,23 +23,36 @@ type Report struct {
...
 	return
 }
 
 // CreateReportWithContext performs the same operation as CreateReport, but
 // allows specifying a context that can interrupt the request.
-func (c *Client) CreateReportWithContext(ctx context.Context, typ string, in *Report) (out *Report, err error) {
-	err = c.post(ctx, "reports/"+typ, in, &out)
+func (c *Client) CreateReportWithContext(ctx context.Context, typ string, in *Report, opts *CreateReportOptions) (out *Report, err error) {
```
Justintime50, line 53: `typ` should be `type`, I know you didn't do it but let's fix it.

Justintime50 (changes requested):
Let's do a tad more research, is it possible we can conditionally check for the absence of the second options param so we don't break people's integrations?

report.go
nwithan8: `type` is a reserved word in Golang

report.go
nwithan8: `type` is a reserved word in Golang

Justintime50: Le sigh. Fine

nwithan8:
Putting in draft mode due to breaking change, requires further discussion

nwithan8:
Refactor to include columns in the JSON body rather than as query parameters no longer introduces a breaking change to the `CreateReport` function's signature.

tests/report_test.go
```diff
@@ -87,6 +87,38 @@ func (c *ClientTests) TestReportTrackerCreate() {
...
+
+	report, _ := client.CreateReport(
+		"shipment",
+		&easypost.Report{
+			StartDate: c.fixture.ReportStartDate(),
+			EndDate:   c.fixture.ReportEndDate(),
+			Columns:   []string{"usps_zone"},
+		},
+	)
+
```
jchen293, line 102: Add a comment here to explain why there is only one assertion, similar to what you have done on the other libs.

tests/report_test.go
```diff
@@ -87,6 +87,38 @@ func (c *ClientTests) TestReportTrackerCreate() {
...
+	report, _ := client.CreateReport(
+		"shipment",
+		&easypost.Report{
+			StartDate:         c.fixture.ReportStartDate(),
+			EndDate:           c.fixture.ReportEndDate(),
+			AdditionalColumns: []string{"from_name", "from_company"},
+		},
+	)
+
```
jchen293, line 118: Add a comment here to explain why there is only one assertion, similar to what you have done on the other libs.

jchen293 (commented):
code looks good, just need to add comment in the unit tests

tests/report_test.go
nwithan8: Resolved in [ee78831](https://github.com/EasyPost/easypost-go/pull/63/commits/ee7883187e310394a3eabb3df65c7825e06effac)

tests/report_test.go
nwithan8: Resolved in [ee78831](https://github.com/EasyPost/easypost-go/pull/63/commits/ee7883187e310394a3eabb3df65c7825e06effac)

jchen293 (approved):

PR #81, opened 2022-05-24
feat: scrub sensitive data from cassettes

jchen293:
# Description

Remove sensitive data from the cassettes. 
These changes (at this point) do not scrub nested keys and only scrub at the root level.
New test coverage report
Screen Shot 2022-05-24 at 6 05 24 PM

# Testing

Re-recorded cassettes that previously had data that is now scrubbed (was not sensitive as it was dummy data)
Stopped skipping previous tests now that scrubbing can safely remove the sensitive data

# Pull Request Type

- [x] Improvement (fixing a typo, updating readme, renaming a variable name, etc)


tests/util_test.go
```diff
@@ -55,8 +57,68 @@ func (c *ClientTests) SetupTest() {
...
+		for scrubber, replaceValue := range responseBodyScrubbers {
+			if strings.Contains(i.Response.Body, scrubber) {
+				var responseBody = i.Response.Body
+
+				if responseBody[0] == '[' && responseBody[len(responseBody)-1] == ']' {
+					var arrayObjects []map[string]interface{}
+
+					err := json.Unmarshal([]byte(responseBody), &arrayObjects)
+					if err != nil {
+						panic(err)
```
Justintime50, line 93: Let's use `os.Exit` here instead of `panic`: https://stackoverflow.com/a/28473339/6064135 (make sure to print the error message prior to using os.Exit.)

This goes for all the panics in this PR (you'll notice we don't use Panic elsewhere in the lib).

tests/util_test.go
```diff
@@ -55,8 +57,68 @@ func (c *ClientTests) SetupTest() {
...
+
+					for index, object := range arrayObjects {
+						for key := range object {
+							if key == scrubber {
+								arrayObjects[index][key] = replaceValue
+							}
+						}
+					}
+
+					responseBodyByte, _ = json.Marshal(arrayObjects)
```
Justintime50, line 104: Should be plural because there's lots of bytes in there, not just one.

```suggestion
					responseBodyBytes, _ = json.Marshal(arrayObjects)
```

Justintime50 (changes requested):
I'd like to know what the new test coverage report says now that we've added these few missing tests in addition to fixing the few items I've pointed out here.

Justintime50 (commented):
If we are properly scrubbing, we can also now unskip the API key unit tests (please double/triple check the cassettes before committing)

jchen293:
> If we are properly scrubbing, we can also now unskip the API key unit tests (please double/triple check the cassettes before committing)

It's unskipped, Go lib does not have this Python function where it `Get the authenticated user's API keys.`: https://github.com/EasyPost/easypost-python/blob/master/easypost/user.py#L59

The cassette is added.

tests/util_test.go
jchen293: resolved

tests/util_test.go
```diff
@@ -55,8 +57,68 @@ func (c *ClientTests) SetupTest() {
...
+
+		// Filter response bodies
+		var responseBodyByte []byte
+		var responseBodyString string
+
+		for scrubber, replaceValue := range responseBodyScrubbers {
+			if strings.Contains(i.Response.Body, scrubber) {
+				var responseBody = i.Response.Body
+
+				if responseBody[0] == '[' && responseBody[len(responseBody)-1] == ']' {
```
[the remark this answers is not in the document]
jchen293: logic has been changed

jchen293:
> I'd like to know what the new test coverage report says now that we've added these few missing tests in addition to fixing the few items I've pointed out here.

I have uploaded the test coverage report

tests/util_test.go
```diff
@@ -55,8 +57,80 @@ func (c *ClientTests) SetupTest() {
...
+				} else {
+					responseMap := make(map[string]interface{})
+
+					err := json.Unmarshal([]byte(responseBody), &responseMap)
+					if err != nil {
+						fmt.Print(err)
+						os.Exit(1)
+					}
+
+					if _, exists := responseMap[scrubber]; exists {
```
Justintime50, line 123: Explicit over implicit. Your future self (and other engineers) will thank you!

```suggestion
					if _, keyExists := responseMap[scrubber]; exists {
```

tests/cassettes/TestAddressCreate.yaml
```diff
@@ -6,27 +6,29 @@ interactions:
...
       User-Agent:
-      - EasyPost/v2 GoClient/2.0.1 Go/go1.17.5 OS/darwin
+      - REDACTED
     url: https://api.easypost.com/v2/addresses
     method: POST
   response:
-    body: '{"id":"adr_5408ef399b2511ec9bc7ac1f6b0a0d1e","object":"Address","created_at":"2022-03-03T19:08:36+00:00","updated_at":"2022-03-03T19:08:36+00:00","name":"Jack
-      Sparrow","company":"EasyPost","street1":"388 Townsend St","street2":"Apt 20","city":"San
-      Francisco","state":"CA","zip":"94107","country":"US","phone":"5555555555","email":null,"mode":"test","carrier_facility":null,"residential":null,"federal_tax_id":null,"state_tax_id":null,"verifications":{}}'
+    body: '{"carrier_facility":null,"city":"San Francisco","company":"EasyPost","country":"US","created_at":"2022-05-24T22:28:07+00:00","email":"REDACTED","federal_tax_id":null,"id":"adr_c9060649dbb011ec8b29ac1f6bc72124","mode":"test","name":"Jack
```
Justintime50, line 18: Hmm... because we are converting to a string and back again, the ordering is getting all messed up. In a perfect world, the keys would stay ordered in the order they were defined in the struct (old change on the left) vs now they appear to be ordered alphabetically. Is there a way we can preserve the old ordering?

Justintime50 (changes requested):

tests/cassettes/TestAddressCreate.yaml
jchen293: `json.Unmarshal` converts the json string to a map with random order every time because the map is unordered in Golang, and `json.Marshal` convert map to a json string in alphabetical order

Yes, it's possible if we implement the `json.Marshaler` interface on a custom type. Code has been added and re-recorded cassettes.

tests/cassettes/TestCarrierAccountAll.yaml
```diff
@@ -5,51 +5,26 @@ interactions:
...
-      Account Number","value":"A1A1A1"}}},"credentials":{"account_number":"A1A1A1"},"test_credentials":null},{"id":"ca_9bf8286878e8466ba03a563a5f2db15a","object":"CarrierAccount","type":"UpsAccount","clone":false,"created_at":"2022-03-03T02:44:49Z","updated_at":"2022-03-03T02:44:49Z","description":"UPS
-      Account","reference":null,"readable":"UPS","logo":null,"fields":{"credentials":{"account_number":{"visibility":"readonly","label":"UPS
-      Account Number","value":"A1A1A1"}}},"credentials":{"account_number":"A1A1A1"},"test_credentials":null},{"id":"ca_0ff1a988c20442b5aa6246d4356ecb99","object":"CarrierAccount","type":"UpsAccount","clone":false,"created_at":"2022-03-03T02:44:49Z","updated_at":"2022-03-03T02:44:50Z","description":"My
-      custom description","reference":null,"readable":"UPS","logo":null,"fields":{"credentials":{"account_number":{"visibility":"readonly","label":"UPS
-      Account Number","value":"A1A1A1"}}},"credentials":{"account_number":"A1A1A1"},"test_credentials":null}]'
+    body: '[{"billing_type":"easypost","clone":false,"created_at":"2020-05-07T23:43:48Z","credentials":{},"description":"USPS
```
Justintime50, line 15: It appears that nested objects in a list do not have the correct key ordering here.

tests/cassettes/TestInsuranceAll.yaml
```diff
@@ -5,109 +5,116 @@ interactions:
...
-      TX, 77001","origin_tracking_location":{"object":"TrackingLocation","city":"NORTH
-      HOUSTON","state":"TX","country":null,"zip":"77315"},"destination_location":"CHARLESTON
-      SC, 29401","destination_tracking_location":{"object":"TrackingLocation","city":"CHARLESTON","state":"SC","country":null,"zip":"29407"},"guaranteed_delivery_date":null,"alternate_identifier":null,"initial_delivery_attempt":"2022-02-05T16:32:13Z"},"finalized":true,"is_return":false,"public_url":"https://track.easypost.com/djE6dHJrXzNlODNlYmU2NGZiYTQ3YmRhMjdiZjYwODkzZTQ4NmI3","fees":[]},"tracking_code":"9400100109361109729572","fee":{"object":"Fee","type":"InsuranceFee","amount":"1.00000","charged":true,"refunded":false},"messages":["The
-      tracker entered the mailstream before insurance was purchased."],"created_at":"2022-03-03T02:38:12Z","updated_at":"2022-03-03T02:39:45Z"}],"has_more":true}'
+    body: '{"insurances":[{"amount":"100.00000","created_at":"2022-05-25T18:03:30Z","fee":{"amount":"1.00000","charged":true,"object":"Fee","refunded":false,"type":"InsuranceFee"},"from_address":{"carrier_facility":null,"city":"San
```
Justintime50, line 17: Yeah, nested objects are definitely out of order.

tests/apikey_test.go
```diff
@@ -0,0 +1,14 @@
+package easypost_test
```
Justintime50, line 1: This file should be `api_key_test.go`

tests/apikey_test.go
```diff
@@ -0,0 +1,14 @@
...
+
+func (c *ClientTests) TestUserApiKeysAll() {
+	client := c.ProdClient()
+	assert := c.Assert()
+
+	apiKeys, err := client.GetAPIKeys()
+	if err != nil {
+		c.T().Error(err)
+		return
+	}
```
Justintime50, line 11: Please look at the new style for checking no error occurred (my last PR for this lib) and use that here instead.

tests/apikey_test.go
```diff
@@ -0,0 +1,14 @@
+package easypost_test
+
+func (c *ClientTests) TestUserApiKeysAll() {
```
Justintime50, line 3: Re-record this cassette please so it picks up the new name.

```suggestion
func (c *ClientTests) TestApiKeysAll() {
```

tests/util_test.go
```diff
@@ -36,13 +38,65 @@ type ClientTests struct {
 	fixture  *Fixture
 }
 
+type OrderedMap struct {
+	Order []string
+	Map   map[string]interface{}
+}
+
+func (om *OrderedMap) UnmarshalJSON(b []byte) error {
```
Justintime50, line 46: Please leave a comment as to why we are unmarshaling and remarshaling above each of these two functions.

Also, see my other comments. I think this only works at the root level. Nested objects are still out of order. If there isn't a way to fix them all, then I'd prefer we don't fix half of them only.

Justintime50 (changes requested):

tests/cassettes/TestCarrierAccountAll.yaml
jchen293: correct, this is only works with non-array objects string.

tests/cassettes/TestInsuranceAll.yaml
jchen293: correct, this only works with non-array object string.

tests/util_test.go
jchen293: These two functions only work with non-array object strings. 

For example, it works when the JSON string is `{id:"xxx", address:"xxx"}` but not in `[{id:"xxx"}, {id:"xxx"}].

Basically, it will not work when the API endpoint return an array object string which happens in `TestCarrierAccountAll`

tests/cassettes/TestInsuranceAll.yaml
Justintime50: Hmm... that's too bad. Thank you for looking into this. If it will only work for one and not the other, I'd prefer if we didn't introduce the complexity and simply kept them all the same (ordered alphabetically). I'll cry myself to sleep but we shouldn't spend more time on it.

tests/util_test.go
nwithan8: It looks like json.Marshal should be able to handle arrays as well as dictionaries, or am I interpreting this wrong? https://golang.cafe/blog/golang-json-marshal-example.html

nwithan8 (dismissed):

tests/util_test.go
jchen293: Yes, `json.Marshal` can handle both but it will sort the items in alphabetical order which is not what we wanted.

jchen293: We have decided to remove this function and keep the order alphabetical

Justintime50 (approved):
Nice work! Thanks for doing the extra work to investigate cassette ordering - even though we didn't use that piece here, you were able to do a bit of learning 🙂

tests/util_test.go
```diff
@@ -55,8 +57,77 @@ func (c *ClientTests) SetupTest() {
...
+		"credentials":      map[string]string{},
+		"email":            "REDACTED",
+		"keys":             []*easypost.APIKey{},
+		"key":              "REDACTED",
+		"phone_number":     "REDACTED",
+		"phone":            "REDACTED",
+	}
+	r.AddSaveFilter(func(i *cassette.Interaction) error {
+		// Filter headers
+		if i.Request.Headers["Authorization"] != nil {
```
nwithan8, line 76: nit: We could have made this a loop of header strings.

nwithan8 (approved):
LGTM, great job on this!

PR #84, opened 2022-05-31
Swapped PUT with PATCH

ryannguyen16:
# Description

- Swapped update brand, carrier, user, and webhook endpoints to use patch instead

# Testing

# Pull Request Type

- [x] Improvement (fixing a typo, updating readme, renaming a variable name, etc)

### DIFF
--- a/tests/carrier_account_test.go
+++ b/tests/carrier_account_test.go
@@ -27,7 +27,6 @@ func (c *ClientTests) TestCarrierAccountCreate() {
 	assert.Equal("UpsAccount", carrierAccount.Type)
 
 	err = client.DeleteCarrierAccount(carrierAccount.ID)
-
 	require.NoError(err)
 }
 
@@ -45,7 +44,6 @@ func (c *ClientTests) TestCarrierAccountRetrieve() {
 	assert.Equal(carrierAccount, retrievedCarrierAccount)
 
 	err = client.DeleteCarrierAccount(carrierAccount.ID)
-
 	require.NoError(err)
 }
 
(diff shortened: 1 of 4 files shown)


nwithan8:
Please remember to follow the PR template (fill out the "Description" and "Testing" sections and check the proper box in "Pull Request Type"), and then add @nwithan8 , @jchen293  and @Justintime50 as "Reviewers" on the top of the right-hand column.

nwithan8:
I'm not sure why the GitHub Actions isn't finding this cassette. There is for sure a PATCH call to this URL in the cassettes that Ryan included: https://github.com/EasyPost/easypost-go/runs/6676928709?check_suite_focus=true#step:5:17

Justintime50:
@nwithan8 I believe you're looking at a red herring. 

@ryannguyen16, you will want to update the `TestCarrierAccountUpdate` test to assert that the name of the user got updated instead of the phone number. It's currently failing because we redact the phone number in the cassettes but didn't re-record this cassette previously. You can look at the Ruby test as an example of how you can change this one: https://github.com/EasyPost/easypost-ruby/blob/261b0f409961a087fb38f4d534d20aa8911f2560/spec/user_spec.rb#L40-L51 (note this will update your user's name. You may want to go back and manually update it in the Dashboard once done)

Let me know if you have questions!

ryannguyen16:
@nwithan8 Sorry about that. I'll make sure to have the right format for next time.

Justintime50:
@nwithan8 so this one's interesting. It's actually failing on a couple items (mostly a VCR problem). The test run needs fixing for the update user test for sure to assert against a name instead of the phone number (easily fixed), but even doing that, it cannot replay the update carrier account test properly due to a matching error.

Justintime50:
For now, I've skipped that test and created a task to revisit. Looks like there are a couple tests across each lib that can't match cassettes due to scrubbing that we should circle back to.

jchen293 (approved):

PR #91, opened 2022-06-16
Updated withContext functions.

ryannguyen16:
# Description

Changed the withContext functions in the go library so it can be covered by the tests.

# Testing

Screen Shot 2022-06-16 at 10 28 10 AM

# Pull Request Type

### DIFF
--- a/address.go
+++ b/address.go
@@ -111,9 +111,7 @@ type verifyAddressResponse struct {
 //		&CreateAddressOptions{Verify: []string{"delivery"}},
 //	)
 func (c *Client) CreateAddress(in *Address, opts *CreateAddressOptions) (out *Address, err error) {
-	req := &createAddressRequest{CreateAddressOptions: opts, Address: in}
-	err = c.post(context.Background(), "addresses", req, &out)
-	return
+	return	c.CreateAddressWithContext(nil, in, opts)
 }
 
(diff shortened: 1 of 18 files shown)


Justintime50:
@ryannguyen16 you'll want to fix the lint errors here. If you run `make lint` it will tell you what is wrong that needs fixing.

nwithan8:
> @ryannguyen16 you'll want to fix the lint errors here. If you run `make lint` it will tell you what is wrong that needs fixing.

Um, the errors are because we're passing `nil` context into these functions, which ... is the whole point of this PR...
https://github.com/EasyPost/easypost-go/pull/91/files

nwithan8 (approved):

Justintime50 (approved):

qeubar:
> > Um, the errors are because we're passing `nil` context into these functions, which ... is the whole point of this PR... https://github.com/EasyPost/easypost-go/pull/91/files
> 
> Nah, instead of passing `nil` as the context, pass `context.Background()` and the linter will stop complaining.

that was a typo on my original comment from the past. the context var should never be nil, `context.Background()` or `context.TODO()` should be used. https://pkg.go.dev/context

qeubar (approved):

PR #93, opened 2022-06-27
Add lowest rate/smartrate support

nwithan8:
# Description

- Add improved functions for getting lowest rate for order, shipment and pickup
- Deprecate old functions
- Add functions for getting lowest smartrate for shipment
- Add unit tests

# Testing

- Re-record all cassettes due to USPS hiccup
- All tests pass

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Adds `BillingType` in CarrierAccount and Rate structs
 - Add new lowest rate functions for Shipments, Orders and Pickups
 - Add new lowest smartrate functions for Shipments
+- Adds OS details to the User-Agent header
 
 ## v2.3.0 (2022-04-13)
 
--- a/client.go
+++ b/client.go
@@ -24,9 +24,10 @@ var defaultUserAgent string
 var defaultTimeout int
 
 func init() {
+	// We skip grabbing the OS version (for now) as there is not a reliable way to do so across OS's
 	defaultUserAgent = fmt.Sprintf(
-		"EasyPost/v2 GoClient/%s Go/%s OS/%s",
-		Version, runtime.Version(), runtime.GOOS)
+		"EasyPost/v2 GoClient/%s Go/%s OS/%s OSVersion/%s OSArch/%s",
+		Version, runtime.Version(), runtime.GOOS, "NA", runtime.GOARCH)
 	defaultTimeout = 60000
 }
 


jchen293:
why was every cassette re-recorded?

util.go
```diff
@@ -9,3 +16,170 @@ func StringPtr(s string) *string {
...
+	// if not rate was ever set (nothing matched the criteria), return an error
+	if (out == RateFilterRate{}) {
+		return out, errors.New("no rates found")
+	}
+
+	return
+}
+
+// lowestRateImpl returns the lowest rate from the given list of rates.
+func (c *Client) lowestRateImpl(rates []*Rate, carriers []string, services []string) (out Rate, err error) {
```
jchen293, line 142: I think this function should be renamed or spell out `Impl`, perhaps something similar to Java lib `getLowestObjectRate`

tests/payment_methods_test.go
```diff
@@ -8,5 +8,4 @@ func (c *ClientTests) TestPaymentMethodsAll() {
 	require.NoError(err)
 
 	assert.True(paymentMethods.PrimaryPaymentMethod != nil)
-	assert.True(paymentMethods.SecondaryPaymentMethod != nil)
```
jchen293, line 11: I am not sure why this assertion is removed from this PR?

util.go
```diff
@@ -9,3 +16,170 @@ func StringPtr(s string) *string {
...
+
+// containsString returns true if the given list contains the given element.
+func containsString(s []string, e string) bool {
+	for _, a := range s {
+		if a == e {
+			return true
+		}
+	}
+	return false
+}
```
jchen293, line 28: Using `a` and `s` as the variable names is quite confusing for the future reader, can you rename them?

util.go
```diff
@@ -9,3 +16,170 @@ func StringPtr(s string) *string {
...
+			if rate.ID == lowestRate.ID {
+				return *rate, nil
+			}
+		}
+	}
+	return
+}
+
+// lowestPickupRateImpl returns the lowest pickup rate from the given list of pickup rates.
+func (c *Client) lowestPickupRateImpl(rates []*PickupRate, carriers []string, services []string) (out PickupRate, err error) {
```
jchen293, line 165: See above comment about the function name

jchen293 (commented):
why was every cassette re-recorded?

tests/payment_methods_test.go
nwithan8: The cassette does not have a secondary payment method in it.

Justintime50: The purpose of these assertions is to ensure the keys exist. We may need to adjust it to account for keys only and not values like we did with one of the other libs.

util.go
nwithan8: Resolved

tests/payment_methods_test.go
nwithan8: Re-added line, tweaked a bit to account for the cassette

Justintime50 (commented):
We should only be re-recording cassettes that needed it, not the whole test suite heh.

shipment.go
```diff
@@ -326,47 +323,45 @@ func (c *Client) RerateShipmentWithContext(ctx context.Context, shipmentID strin
...
-	return c.LowestRateWithCarrier(shipment, nil)
+	return c.LowestShipmentRate(shipment)
 }
 
+// LowestShipmentRate gets the lowest rate of a shipment
+func (c *Client) LowestShipmentRate(shipment *Shipment) (out Rate, err error) {
+	return c.LowestShipmentRateWithCarrier(shipment, nil)
+}
+
+// Deprecated: Use LowestShipmentRateWithCarrier instead.
```
Justintime50, line 337: Question: Have we confirmed the godoc pics up on this correctly or do we instead need to have this be the 2nd line in the summary?

shipment.go
```diff
@@ -326,47 +323,45 @@ func (c *Client) RerateShipmentWithContext(ctx context.Context, shipmentID strin
...
+	return c.LowestShipmentRateWithCarrier(shipment, carriers)
 }
 
+// LowestShipmentRateWithCarrier performs the same operation as LowestShipmentRate,
+// but allows specifying a list of carriers for the lowest rate
+func (c *Client) LowestShipmentRateWithCarrier(shipment *Shipment, carriers []string) (out Rate, err error) {
+	return c.LowestShipmentRateWithCarrierAndService(shipment, carriers, nil)
+}
+
+// Deprecated: Use LowestShipmentRateWithCarrierAndService instead.
```
Justintime50, line 350: Question: Have we confirmed the godoc pics up on this correctly or do we instead need to have this be the 2nd line in the summary?

util.go
```diff
@@ -9,3 +16,170 @@ func StringPtr(s string) *string {
 func BoolPtr(b bool) *bool {
 	return &b
 }
+
+// containsString returns true if the given list contains the given element.
+func containsString(strings []string, targetString string) bool {
```
Justintime50, line 21: suggestion: it may be good to clarify what contains string in the function name. `listContainsString` maybe?

util.go
```diff
@@ -9,3 +16,170 @@ func StringPtr(s string) *string {
...
+	for _, aString := range strings {
+		if aString == targetString {
+			return true
+		}
+	}
+	return false
+}
+
+type RateFilterRate struct {
```
Justintime50, line 30: issue: can we rename this? Why do we repeat `Rate` in the name? We should be more descriptive.

Additionally, this needs to be at the top of the file. Structs come before functions.

Justintime50 (dismissed):
Couple small things, otherwise looks good.

shipment.go
nwithan8: https://github.com/golang/go/wiki/Deprecated#:~:text=The%20paragraph%20does%20not%20have%20to%20be%20the%20last%20paragraph%20in%20the%20doc%20comment.

util.go
nwithan8: It's a minified rate that goes in the rate filter. It's a rate-filter rate :)
Any better suggestions?

nwithan8: Resolved

util.go
```diff
@@ -9,3 +23,163 @@ func StringPtr(s string) *string {
...
+	return false
+}
+
+// lowestSmartRate returns the lowest smartrate from the given list of smartrates.
+func (c *Client) lowestSmartRate(rates []*SmartRate, deliveryDays int, deliveryAccuracy string) (out SmartRate, err error) {
+	validDeliveryAccuracies := []string{"percentile_50", "percentile_75", "percentile_85", "percentile_90", "percentile_95",
+		"percentile_97", "percentile_99"}
+
+	// if the delivery accuracy is not valid, return an error
+	if !listContainsString(validDeliveryAccuracies, deliveryAccuracy) {
```
jchen293, line 43:
```suggestion
	if !listContainsString(validDeliveryAccuracies, strings.ToLower(deliveryAccuracy)) {
```

Let's lowercase `deliberyAccuracy` so it will not be case sensitive when checking `deliveryAccuracy` is valid or not.

util.go
Justintime50: Your alternative name used is great, thanks

Justintime50 (approved):

util.go
```diff
@@ -9,3 +23,163 @@ func StringPtr(s string) *string {
...
+
+	// if the delivery accuracy is not valid, return an error
+	if !listContainsString(validDeliveryAccuracies, strings.ToLower(deliveryAccuracy)) {
+		return out, fmt.Errorf("invalid delivery accuracy: %s", deliveryAccuracy)
+	}
+
+	for _, rate := range rates {
+		smartrateDeliveryDay := -1
+
+		switch deliveryAccuracy {
```
jchen293, line 50:
```suggestion
		switch strings.ToLower(deliveryAccuracy) {
```
This one needs to be lowercased as well, then I will stamp.

jchen293 (approved):

jchen293 (changes requested):
Actually, you need to update the changelog for this PR.

PR #94, opened 2022-06-28
chore: ignore cassette diffs

Justintime50:
# Description

Ignore cassette diffs locally (line 1) and collapse on GitHub (line 2) for easier review.

# Pull Request Type

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+* text=auto
+
+tests/cassettes/* -diff
+tests/cassettes/* linguist-generated
--- a/tests/cassettes/TestAddressAll.yaml
+++ b/tests/cassettes/TestAddressAll.yaml
@@ -14,28 +14,28 @@ interactions:
     url: https://api.easypost.com/v2/addresses
     method: GET
   response:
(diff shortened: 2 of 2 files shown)


nwithan8:
Diff stats for this PR still reflect the changes made to the cassette...

.gitattributes
```diff
@@ -0,0 +1,4 @@
+* text=auto
```
nwithan8, line 1: See comment above.

nwithan8 (changes requested):

.gitattributes
Justintime50: You cannot remove the lines added/removed via this method. The "stats" that GitHub's docs talk about is the language breakdown. They built `linguist` to determine the language makeup of a repo, so cassettes won't count towards that. But the bigger thing this solves is that now cassettes will always be collapsed by default and will not show the actual diff on GitHub unless you open up the file.

nwithan8: Bummer, so still big green numbers. ¯\_(ツ)_/¯

nwithan8 (approved):

PR #115, opened 2022-08-18
Add EndShipper

jchen293:
# Description

Adds the new `EndShipper` beta class with create, retrieve, all, and update functions. This is the fully-qualified address object that the USPS will start requiring shortly.

# Testing

Add unit tests and pass them

# Pull Request Type

- [ ] Improvement (fixing a typo, updating readme, renaming a variable name, etc)

### DIFF
--- a/endshipper.go
+++ b/endshipper.go
@@ -1,0 +1,71 @@
+package easypost
+
+import (
+	"context"
+	"net/http"
+)
+
+// ListEndShipperResult holds the results from the list EndShippers API.
+type ListEndShipperResult struct {
+	EndShippers []*Address `json:"endshippers,omitempty"`
+	// HasMore indicates if there are more responses to be fetched. If True,
+	// additional responses can be fetched by updating the ListEndShipperOptions
+	// parameter's AfterID field with the ID of the last item in this object's
+	// EndShippers field.
+	HasMore bool `json:"has_more,omitempty"`
+}
+
+// CreateEndShipper submits a request to create a new endshipper, and returns the result.
+func (c *Client) CreateEndShipper(in *Address) (out *Address, err error) {
+	return c.CreateEndShipperWithContext(context.Background(), in)
+}
+
+// CreateEndShipperWithContext performs the same operation as CreateEndShipper, but
+// allows specifying a context that can interrupt the request.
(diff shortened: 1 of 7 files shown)


Justintime50 (approved):

endshipper.go
```diff
@@ -0,0 +1,71 @@
+package easypost
+
+import (
+	"context"
+	"net/http"
+)
+
+// ListEndShipperResult holds the results from the list EndShippers API.
+type ListEndShipperResult struct {
+	EndShippers []*Address `json:"endshippers,omitempty"`
```
nwithan8, line 10: Do we not want to make a specific class for EndShipper? I know it's basically just an Address based on the properties, but for the future? Also to prevent someone from using an EndShipper as a normal address?

nwithan8 (changes requested):

endshipper.go
jchen293: Per internal discussion, the answer is we are going to `address` for now. cc @Justintime50

nwithan8 (approved):

endshipper.go
Justintime50: Well, why not split them up? What did we do with Java/C#?

Justintime50 (changes requested):
See my last comment

endshipper.go
jchen293: We use inheritance in C# and Java, but unfortunately in Go there is no Inheritance

Justintime50: Hmm.. what should we do? Run the risk of objects changing over time or reuse Address because EndShipper is literally an Address?

jchen293: I agree with Nate, if somehow in the future EndShipper properties diverge from Address, then it would be better to have EndShipper struct itself

nwithan8: I say diverge and make an explicit class.

I know I was a big proponent of base classes and inheritance for C# and Java, but honestly, Address has been the only thing so far that we've used it for, so it's largely been unnecessary.

Justintime50: Let's split em up then!

Justintime50 (approved):

endshipper.go
jchen293: We have decided not to split em up

PR #126, opened 2022-09-21
Custom (un)marshalling for CustomsItem to account for Value type

nwithan8:
# Description

- Custom (un)marshal functions for CustomsItem to account for float/string discrepancy with Value
  - Value will always be a string when serialized to JSON
  - Value can be either a string or a float when deserializing from JSON
- Remove workaround from fixtures
- Re-record cassettes as needed

# Testing

- Manually tested and confirmed that deserializing a string (via API response) and a float (via fixture JSON) both work
- Cassettes re-recorded as needed
- All unit tests pass

# Pull Request Type


customs.go
```diff
@@ -47,6 +50,70 @@ type createCustomsItemRequest struct {
...
+		case "value":
+			// convert Value to a float64 if it is a string (via API response)
+			if s, ok := v.(string); ok {
+				if f, err := strconv.ParseFloat(s, 64); err == nil {
+					rawData[k] = f
+				}
+			} else {
+				// keep Value a float64 if it is already one (via Fixture data)
+				rawData[k] = v.(float64)
+			}
```
jchen293, line 93:
```suggestion
	for key, value := range rawData {
		switch key {
		case "value":
			// convert Value to a float64 if it is a string (via API response)
			if stringValue, noErr := value.(string); noErr {
				if floatValue, err := strconv.ParseFloat(stringValue, 64); err == nil {
					rawData[key] = floatValue
				}
			} else {
				// keep Value a float64 if it is already one (via Fixture data)
				rawData[key] = value.(float64)
			}
```

Nit: improve the readability by expliciting the single character variable name.
Question: What if the `strconv.ParseFloat(s, 64)` failed or has error?

nwithan8: Resolved.

If an error occurs internally, the method will return (an error)

PR #134, opened 2022-11-22
Use different endpoint for custom workflow carrier account creation (FedEx, UPS)

nwithan8:
# Description

- Carrier account creation now determines endpoint based on type

# Testing

- New unit test, cassette
- Re-record related cassettes (request body changed)

# Pull Request Type

### DIFF
--- a/carrier.go
+++ b/carrier.go
@@ -53,9 +53,7 @@ type carrierAccountRequest struct {
 }
 
 func (c *Client) selectCarrierAccountCreationEndpoint(account CarrierAccount) string {
-	carriersWithCustomWorkflow := []string{"FedexAccount", "UpsAccount"}
-
-	for _, carrier := range carriersWithCustomWorkflow {
+	for _, carrier := range getCustomWorkflowCarrierAccountTypes() {
 		if account.Type == carrier {
 			return "carrier_accounts/register"
 		}
--- a/carrier.go
+++ b/carrier.go
@@ -53,7 +53,7 @@ type carrierAccountRequest struct {
 }
 
 func (c *Client) selectCarrierAccountCreationEndpoint(account CarrierAccount) string {
-	for _, carrier := range getCustomWorkflowCarrierAccountTypes() {
+	for _, carrier := range getCarrierAccountTypesWithCustomWorkflows() {
 		if account.Type == carrier {
 			return "carrier_accounts/register"
 		}
--- a/carrier.go
+++ b/carrier.go
@@ -24,19 +24,20 @@ type CarrierFields struct {
 // CarrierAccount encapsulates credentials and other information related to a
 // carrier account.
 type CarrierAccount struct {
-	ID              string            `json:"id,omitempty"`
-	Object          string            `json:"object,omitempty"`
-	Reference       string            `json:"reference,omitempty"`
-	CreatedAt       *time.Time        `json:"created_at,omitempty"`
-	UpdatedAt       *time.Time        `json:"updated_at,omitempty"`
-	Type            string            `json:"type,omitempty"`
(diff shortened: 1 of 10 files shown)


carrier.go
```diff
@@ -24,19 +24,20 @@ type CarrierFields struct {
...
+	CreatedAt        *time.Time             `json:"created_at,omitempty"`
+	UpdatedAt        *time.Time             `json:"updated_at,omitempty"`
+	Type             string                 `json:"type,omitempty"`
+	Fields           *CarrierFields         `json:"fields,omitempty"`
+	Clone            bool                   `json:"clone,omitempty"`
+	Description      string                 `json:"description,omitempty"`
+	Readable         string                 `json:"readable,omitempty"`
+	Credentials      map[string]string      `json:"credentials"`
+	TestCredentials  map[string]string      `json:"test_credentials"`
+	RegistrationData map[string]interface{} `json:"registration_data"`
```
Justintime50, line 39: These all need `,omitempty` added since it could be either or and some may have prod or test creds and not the other.

tests/carrier_account_test.go
```diff
@@ -15,6 +15,18 @@ func (c *ClientTests) GenerateCarrierAccount() (*easypost.CarrierAccount, error)
 	return carrierAccount, err
 }
 
+func (c *ClientTests) GenerateCarrierAccountWithCustomWorkflow() (*easypost.CarrierAccount, error) {
```
Justintime50, line 18: Having these test helper functions aren't actually very helpful (especially for one-offs like this). By moving the test logic outside the test itself, it adds more indirection. Can we keep all test logic inside each test so that it's contained?

Now, if this same logic was used in a half dozen tests, it would make sense, but since that's not the case it really only adds more places to look to learn about a test than it provides value. This is true for your other recent PRs that touch tests.

tests/carrier_account_test.go
```diff
@@ -30,6 +42,36 @@ func (c *ClientTests) TestCarrierAccountCreate() {
...
+	// We expect the error to be an API error
+	assert.Equal(reflect.TypeOf(&easypost.APIError{}), reflect.TypeOf(err))
+
+	// We expect the error to be a 422
+	assert.Equal(422, err.(*easypost.APIError).StatusCode)
+
+	// We expect the error to have sub-errors
+	assert.NotEmpty(err.(*easypost.APIError).Errors)
+
+	// We expect one of the sub-errors to be regarding a missing field
```
Justintime50, line 62: I'd argue many of these comments are self-explanatory since the code is documenting itself. It may be better to have a single comment at the top of this test describing what's expected as a whole based on the data passing and why instead of commenting each line individually.

tests/carrier_account_test.go
```diff
@@ -30,6 +42,36 @@ func (c *ClientTests) TestCarrierAccountCreate() {
...
+	errorFound := false
+	for _, err := range err.(*easypost.APIError).Errors {
+		if err.Field == "account_number" && err.Message == "must be present and a string" {
+			errorFound = true
+			break
+		}
+	}
+
+	assert.True(errorFound)
+
```
Justintime50, line 72: Remove this extra line please.

Justintime50 (changes requested):

carrier.go
nwithan8: Does sending a "null" to the API versus not sending it at all make a difference?

Justintime50: It could. In this lib, just about every field has `omitempty` so we should follow convention here.

jchen293 (approved):

Justintime50 (approved):

PR #141, opened 2023-01-09
add stripe refund and add payment beta functionalities

jchen293:
# Description

Beta functions are added in the existing `Referral` file because there is no class and object in Go so we don't really need to create a separate beta file for it. 

- [ADDED] Adds new beta billing functionality for ReferralCustomer users
  - `BetaAddPaymentMethod` can add a pre-existing Stripe bank account or credit card to your EasyPost account
  - `BetaRefundByAmount` refunds your wallet by a dollar amount
  - `BetaRefundByPaymentLog` refunds your wallet by a PaymentLog ID
# Testing

Add unit tests and pass

# Pull Request Type

### DIFF
--- a/referral.go
+++ b/referral.go
@@ -11,6 +11,15 @@ import (
 	"strings"
 )
 
+// A PaymentRefund that has the refund details for the refund request.
+type BetaPaymentRefund struct {
+	RefundedAmount          int        `json:"refundedamount,omitempty"`
+	RefundedPaymentLogs     []string   `json:"refundedpaymentlog,omitempty"`
+	PaymentLogId            string     `json:"paymentlogid,omitempty"`
+	RefundedAmountCurrencys string     `json:"refundedamountcurrencys,omitempty"`
+	Errors                  []APIError `json:"errors,omitempty"`
+}
+
 // A ReferralCustomer contains data about an EasyPost referral customer.
 type ReferralCustomer struct {
 	User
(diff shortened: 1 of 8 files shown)


referral.go
```diff
@@ -187,3 +196,78 @@ func (c *Client) createEasypostCreditCard(ctx context.Context, referralCustomerA
...
+func (c *Client) BetaAddPaymentMethodWithPrimaryOrSecondary(stripeCustomerId string, paymentMethodReference string, primaryOrSecondary PaymentMethodPriority) (out *PaymentMethodObject, err error) {
+	return c.BetaAddPaymentMethodWithPrimaryOrSecondaryAndContext(context.Background(), stripeCustomerId, paymentMethodReference, primaryOrSecondary)
+}
+
+// BetaAddPaymentMethodWithPrimaryOrSecondaryAndContext performs the same operation as BetaAddPaymentMethodWithPrimaryOrSecondary, but allows
+// specifying a context that can interrupt the request.
+func (c *Client) BetaAddPaymentMethodWithPrimaryOrSecondaryAndContext(ctx context.Context, stripeCustomerId string, paymentMethodReference string, primaryOrSecondary PaymentMethodPriority) (out *PaymentMethodObject, err error) {
+	primaryOrSecondaryEndpoint := ""
+
+	switch primaryOrSecondary {
```
nwithan8, line 222: Since we're reusing this primary/secondary switch case in multiple places in our app now, we should consider making this a separate helper function.

referral.go
```diff
@@ -11,6 +11,15 @@ import (
 	"strings"
 )
 
+// A PaymentRefund that has the refund details for the refund request.
+type BetaPaymentRefund struct {
+	RefundedAmount          int        `json:"refundedamount,omitempty"`
+	RefundedPaymentLogs     []string   `json:"refundedpaymentlog,omitempty"`
+	PaymentLogId            string     `json:"paymentlogid,omitempty"`
+	RefundedAmountCurrencys string     `json:"refundedamountcurrencys,omitempty"`
```
Justintime50, line 19: These all need to be snake case json elements. 
```suggestion
	RefundedAmount          int        `json:"refunded_amount,omitempty"`
	RefundedPaymentLogs     []string   `json:"refunded_payment_log,omitempty"`
	PaymentLogId            string     `json:"payment_log_id,omitempty"`
	RefundedAmountCurrencys string     `json:"refunded_amount_currencys,omitempty"`
```

referral.go
```diff
@@ -187,3 +196,78 @@ func (c *Client) createEasypostCreditCard(ctx context.Context, referralCustomerA
...
+}
+
+// BetaRefundByAmountWithContext performs the same operation as BetaRefundByAmount, but allows
+// specifying a context that can interrupt the request.
+func (c *Client) BetaRefundByAmountWithContext(ctx context.Context, refundAmount int) (out *BetaPaymentRefund, err error) {
+	params := map[string]interface{}{
+		"refund_amount": refundAmount,
+	}
+
+	err = c.post(ctx, "/beta/referral_customers/refunds", params, out)
```
nwithan8, line 255: Question: Is this how we do beta endpoints in Go, just add "beta" to the endpoint? No bool or enum that triggers the API version to change? So all other post functions in here say `/v2/something/something`?

referral.go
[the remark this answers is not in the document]
jchen293: Looks like it's not in the response anymore, according to the latest internal doc. We should probably remove it instead?

referral.go
jchen293: Yes, if you pass in the slash in front of the endpoint, it will override the `v2` endpoint. You don't need to append `/v2` in other production functions, we have declared the API version path here: https://github.com/EasyPost/easypost-go/blob/master/client.go#L20

referral.go
```diff
@@ -187,3 +196,78 @@ func (c *Client) createEasypostCreditCard(ctx context.Context, referralCustomerA
...
+// specifying a context that can interrupt the request.
+func (c *Client) BetaAddPaymentMethodWithPrimaryOrSecondaryAndContext(ctx context.Context, stripeCustomerId string, paymentMethodReference string, primaryOrSecondary PaymentMethodPriority) (out *PaymentMethodObject, err error) {
+	primaryOrSecondaryEndpoint := ""
+
+	switch primaryOrSecondary {
+	case PrimaryPaymentMethodPriority:
+		primaryOrSecondaryEndpoint = "primary"
+	case SecondaryPaymentMethodPriority:
+		primaryOrSecondaryEndpoint = "secondary"
+	}
```
Justintime50, line 227: Can this become a helper function since we also have this on the other already-existing functions?

If so, let's replace the other occurrence with the helper function.

referral.go
```diff
@@ -187,3 +196,78 @@ func (c *Client) createEasypostCreditCard(ctx context.Context, referralCustomerA
...
+
+	params := map[string]interface{}{
+		"stripe_customer_id":       stripeCustomerId,
+		"payment_method_reference": paymentMethodReference,
+		"priority":                 primaryOrSecondaryEndpoint,
+	}
+
+	wrappedParams := map[string]interface{}{
+		"payment_method": params,
+	}
```
Justintime50, line 237: Let's eliminate the extra code and just do this inline.

```suggestion
	wrappedParams := map[string]interface{}{
		"payment_method": map[string]interface{}{
		"stripe_customer_id":       stripeCustomerId,
		"payment_method_reference": paymentMethodReference,
		"priority":                 primaryOrSecondaryEndpoint,
	},
	}
```

Justintime50 (changes requested):

referral.go
jchen293: fixed

referral.go
jchen293: A helper function is added

referral.go
jchen293: added

payment_method.go
```diff
@@ -86,3 +86,15 @@ func (c *Client) getPaymentMethodByPriority(priority PaymentMethodPriority) (out
 
 	return out, nil
 }
+
+// GetPaymentEndpointByPrimaryOrSecondary gets the payment priority based on primaryOrSecondary parameter.
+func (c *Client) GetPaymentEndpointByPrimaryOrSecondary(primaryOrSecondary PaymentMethodPriority) (out string) {
```
Justintime50, line 91: We need to use this in the function `getPaymentMethodByPriority` too

Justintime50 (changes requested):

payment_method.go
jchen293: I am not exactly sure what you mean. `getPaymentMethodByPriority` is a separate helper function that has a different return type, it's being used in the `Billing` file

Justintime50: There are two functions that use these enums. Let's use the helper function in both places, I only see it being used in one.

referral.go
nwithan8: Thanks, I hate this anti-pattern 🙃

payment_method.go
jchen293: `getPaymentMethodByPriority ` returns PaymentObject, it's being used in Billing lines 16 and 29: https://github.com/EasyPost/easypost-go/blob/a0a3342fb9ffe7028fd1e4b3924a8a9f5529012b/billing.go#L16

`getPaymentMethodEndpoint` is returning a different endpoint: "credit_cards" or "bank_accounts"

Justintime50: OH. I misunderstood. This doesn't need to be a helper function then since it's only used in one place. This is a bit different than our other libs in that regard because of the enums. Please revert this change.

nwithan8: No wait, it is still being used in a second place:
https://github.com/EasyPost/easypost-go/blob/b590e5b46fe0f2f0c7f40d89aecf46812f7f5f8b/referral.go#L168

nwithan8: This new helper function is used in two places: `createEasypostCreditCard` and `BetaAddPaymentMethodWithPrimaryOrSecondaryAndContext`, it does not need to be reverted.

jchen293: @nwithan8 Thanks for replying while I was finishing up my lunch. Exactly, no need to revert this since the helper function is being used in two places.

nwithan8 (approved):

Issue #156, opened 2023-03-03
[Bug]: SmartRate response failed to unmarshal correctly


tombu-nimble:
### Software Version

2.13.0

### Language Version

1.20

### Operating System

Linux (Alpine under Docker)

### What happened?

1. Create a shipment using an account that has `UPSDAP` enabled, via `CreateShipmentWithContext(context.Context, *Shipment) (*Shipment, error)`
2. Attempt to get smart rates via `GetShipmentSmartratesWithContext(ctx context.Context, shipmentID string) ([]*SmartRate, error)`
3. Got the following error: ` json: cannot unmarshal string into Go struct field SmartRate.result.rate of type float64.`

### What was expected?

Instead of the JSON marshaling error, we should get a list of smart rates back.

### Relevant logs

```sh
Looks like the JSON payload returned from server side mixes floating point numbers and string literals for rates:


  // GET https://api.easypost.com/v2/shipments/shp_redacted/smartrate
  {
    "result": [
      {
        "carrier": "USPS",
        "carrier_account_id": "ca_redacted",
        "created_at": "2023-02-27T23:47:26Z",
        "currency": "USD",
        "delivery_date": null,
        "delivery_date_guaranteed": false,
        "delivery_days": 3,
        "est_delivery_days": 3,
        "id": "rate_redacted",
        "list_currency": "USD",
        "list_rate": 132.18,
        "mode": "test",
        "object": "Rate",
        "rate": 132.18,
        "retail_currency": "USD",
        "retail_rate": 151.4,
        "service": "Priority",
        "shipment_id": "shp_redacted",
        "time_in_transit": {
          "percentile_50": 2,
          "percentile_75": 3,
          "percentile_85": 3,
          "percentile_90": 3,
          "percentile_95": 4,
          "percentile_97": 5,
          "percentile_99": 5
        },
        "updated_at": "2023-02-27T23:47:26Z"
      },
      {
        "carrier": "USPS",
        "carrier_account_id": "ca_redacted",
        "created_at": "2023-02-27T23:47:26Z",
        "currency": "USD",
        "delivery_date": null,
        "delivery_date_guaranteed": false,
        "delivery_days": null,
        "est_delivery_days": null,
        "id": "rate_redacted",
        "list_currency": "USD",
        "list_rate": 319.45,
        "mode": "test",
        "object": "Rate",
        "rate": 319.45,
        "retail_currency": "USD",
        "retail_rate": 360.25,
        "service": "Express",
[...]

PR #157, opened 2023-03-03
Fix type mismatch on rate field of Smart Rates.

tombu-nimble:
# Description

We encountered [issue 156](https://github.com/EasyPost/easypost-go/issues/156) where the server-side response body of Smart Rate API contains mixed string and float literals for rate-related fields. We are an existing customer and feedback have already been provided to customer support. This PR is for sharing what we had to temporarily do to unblock our development, as a reference.

# Pull Request Type

### DIFF
--- a/rate.go
+++ b/rate.go
@@ -2,6 +2,7 @@ package easypost
 
 import (
 	"context"
+	"strconv"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


Justintime50:
@tombu-nimble thank you for submitting this PR! Although this is good work and should do the trick for this edge case with UPS DAP, I believe we want to correct this at the API level (this way it's fixed for everyone, also won't require us to patch this across all our libraries for a single carrier misbehaving). 

The work is scheduled to happen in the next few weeks for UPS DAP. My suggestion is to use these changes locally for now until we have corrected this at the API level. We can revive this PR if that happens to change. For now, I'll mark it as a draft until we know more about how this will get fixed holistically

Justintime50:
Going to close this for now (see thread on the issue for details). If we need this after all, we can reopen at that point.

PR #162, opened 2023-03-28
[feat] Function to get next page of a paginated collection

nwithan8:
# Description

- Add base `PaginatedCollection` struct
- Add private functions to construct `ListOptions` parameters for getting the next page of a paginated collection
- Add public functions for users to get the next page of a specific paginated collection (with or without page size restriction)
- Mark redundant ListReportOptions as deprecated
- Store custom parameters used by report, shipment and tracker lists in resulting structs for re-use during pagination call

# Testing

- Add unit tests, cassettes

# Pull Request Type


address.go
```diff
@@ -134,6 +135,40 @@ func (c *Client) ListAddressesWithContext(ctx context.Context, opts *ListOptions
...
+
+// GetNextAddressPageWithPageSize returns the next page of addresses with a specific page size
+func (c *Client) GetNextAddressPageWithPageSize(collection *ListAddressResult, pageSize int) (out *ListAddressResult, err error) {
+	return c.GetNextPageWithContextWithPageSize(context.Background(), collection, pageSize)
+}
+
+// GetNextPageWithContext performs the same operation as GetNextAddressPage, but
+// allows specifying a context that can interrupt the request.
+func (c *Client) GetNextPageWithContext(ctx context.Context, collection *ListAddressResult) (out *ListAddressResult, err error) {
+	if collection.Addresses == nil || len(collection.Addresses) == 0 {
```
jchen293, line 151: Nit: We could check if `has_more` is `false` upfront here, so you don't need to pass in `collection.HasMore` in the `nextPageParameters` function. If you decide to go this way, `nextPageParameters` and `nextPageParametersWithPageSize` can just be combined into one function, you can pass in `nil` if there is no `pageSize` from the parameter.

error.go
```diff
@@ -46,3 +47,20 @@ func (e *APIError) Error() string {
 	}
 	return fmt.Sprintf("%d %s", e.StatusCode, e.Status)
 }
+
+type LocalError struct {
```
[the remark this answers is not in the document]
nwithan8: Not quite, this is the equivalent of `GeneralError` in C#, something not API-related (the `Error` object versus a true exception is confusing in this lib)

address.go
nwithan8: You can't pass nil to an int variable

address.go
```diff
@@ -86,7 +86,7 @@ type ListAddressResult struct {
 	// additional responses can be fetched by updating the ListAddressOptions
 	// parameter's AfterID field with the ID of the last item in this object's
 	// Addresses field.
-	HasMore bool `json:"has_more,omitempty"`
+	PaginatedCollection
```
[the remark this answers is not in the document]
nwithan8: Maintain it in one place rather than twenty? Part of the new `PaginatedCollection` base struct.

image

address.go
nwithan8: Wait, duh, default 0...

jchen293: If we set it to 0, that would not grab anything, right?

error.go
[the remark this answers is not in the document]
nwithan8: This has been improved

Justintime50 (approved):
Just a reminder to try and be sure to add CHANGELOG entries for everything

Justintime50 (approved):

insurance.go
```diff
@@ -82,6 +79,37 @@ func (c *Client) ListInsurancesWithContext(ctx context.Context, opts *ListOption
 	return
 }
 
+// GetNextInsurancePage returns the next page of addresses
```
jchen293, line 82:
```suggestion
// GetNextInsurancePage returns the next page of insurance.
```

Please double-check all the docstrings in every file to make sure they are correct after copy-paste.

jchen293 (changes requested):

insurance.go
nwithan8: Good catch, thanks!

insurance.go
```diff
@@ -82,6 +79,37 @@ func (c *Client) ListInsurancesWithContext(ctx context.Context, opts *ListOption
...
+// GetNextInsurancePageWithContext performs the same operation as GetNextInsurancePage, but
+// allows specifying a context that can interrupt the request.
+func (c *Client) GetNextInsurancePageWithContext(ctx context.Context, collection *ListInsurancesResult) (out *ListInsurancesResult, err error) {
+	return c.GetNextInsurancePageWithPageSizeWithContext(ctx, collection, 0)
+}
+
+// GetNextInsurancePageWithPageSizeWithContext performs the same operation as GetNextInsurancePageWithPageSize, but
+// allows specifying a context that can interrupt the request.
+func (c *Client) GetNextInsurancePageWithPageSizeWithContext(ctx context.Context, collection *ListInsurancesResult, pageSize int) (out *ListInsurancesResult, err error) {
+	if collection.Insurances == nil || len(collection.Insurances) == 0 {
```
steven-alfonso, line 101:
```suggestion
	if len(collection.Insurances) == 0 {
```

steven-alfonso: `len()` will return `0` if an array/slice is `nil`

list_options.go
```diff
@@ -10,3 +10,19 @@ type ListOptions struct {
 	EndDateTime   *time.Time `url:"end_datetime,omitempty"`
 	PageSize      int        `url:"page_size,omitempty"`
 }
+
+// nextPageParameters returns the next page of a paginated collection.
+// If pageSize is 0, it will use the default page size
+func nextPageParameters(hasMore bool, lastId string, pageSize int) (out *ListOptions, err error) {
```
steven-alfonso, line 16:
```suggestion
func nextPageParameters(hasMore bool, lastID string, pageSize int) (out *ListOptions, err error) {
```

steven-alfonso: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

steven-alfonso: (suggestion) this applies to all instances of using `...Id` instead of `..ID`

list_options.go
```diff
@@ -10,3 +10,19 @@ type ListOptions struct {
...
+		err = EndOfPaginationError
+		return
+	}
+	params := &ListOptions{
+		BeforeID: lastId,
+	}
+	if pageSize > 0 {
+		params.PageSize = pageSize
+	}
+	return params, err
```
steven-alfonso, line 27: this method is using named return params. The first named param is `out`, but is never used

tests/address_test.go
```diff
@@ -74,6 +74,34 @@ func (c *ClientTests) TestAddressAll() {
...
+
+	firstPage, err := client.ListAddresses(
+		&easypost.ListOptions{
+			PageSize: c.fixture.pageSize(),
+		},
+	)
+	require.NoError(err)
+
+	nextPage, err := client.GetNextAddressPageWithPageSize(firstPage, c.fixture.pageSize())
+	defer func() {
```
steven-alfonso, line 89: what is the reason for this `defer()` func? why not place the contents of this function at the end of the outer func?

steven-alfonso (commented):
added some suggestions and questions

insurance.go
nwithan8: Nice, thanks!

tests/address_test.go
nwithan8: Hmm, good point now that I look at it. Was trying to handle the equivalent of a "try-catch" in Go, but looks like I might have wired it redundantly.

Justintime50 (approved):

PR #165, opened 2023-04-13
- Dynamically parse error message when generating a APIError object

nwithan8:
# Description

Most of the time, the `message` property of an `APIError` object's JSON data is a string, but occasionally, our API sends something other than a string.

For example:

"message" is a JSON array
```
{
    "error": {
        ...
        "message": [
                "ERROR 1",
                "ERROR 2"
         ], 
       ...
}
```

"message" is a JSON dictionary
```
{
    "error": {
        ...
        "message": {
            "errors": [
                "ERROR 1",
                "ERROR 2"
            ]
        },
       ...
}
```

This PR enhances the JSON deserialization process for `APIError` objects to account for the "message" being an array, being a JSON dictionary, or being empty.

Shout out to @jchen293 who co-authored this but won't get credit.

# Testing

- Unit tests updated to confirm logic works as expected

# Pull Request Type

- [ ] Improvement (fixing a typo, updating readme, renaming a variable name, etc)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next Release
 
-- The `Message` attribute of an `APIError` is now a `interface{}` rather than a `string`, due to potential inconsistent data structure from the API response
+- The `Message` attribute of an `APIError` is now an `interface{}` rather than a `string`, due to potential inconsistent data structure from the API response
   - Behind-the-scenes, the `message` portion of the JSON response is transformed to a concatenated string. Users should be able to safely cast the `Message` attribute to a string when accessing it via `myApiError.Message.(string)`
 
 ## v2.14.0 (2023-04-04)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## Next Release
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## Next Release
+
+- The `Message` attribute of an `APIError` is now a `interface{}` rather than a `string`, due to potential inconsistent data structure from the API response
```
Justintime50, line 5:
```suggestion
- The `Message` attribute of an `APIError` is now an `interface{}` rather than a `string`, due to potential inconsistent data structure from the API response
```

Justintime50 (commented):
question: This is a breaking change though correct?

nwithan8:
> question: This is a breaking change though correct?

Sort of? Ultimately, `Message` still is a string, and as we've seen in the unit tests, you don't need to explicitly cast it. But technically, it is an `interface{}` according to the compiler, which might interfere with some assumptions elsewhere in a user's application that `Message` is explicitly a `string` type.

jchen293 (approved):
Paired with Nate on this one, it works as expected

Justintime50 (approved):

Issue #183, opened 2023-06-28
[Bug]: GetShipmentEstimatedDeliveryDate fails due to time parsing error when UPS is a carrier


coltmeister:
### Software Version

2.19.0

### Language Version

Golang

### Operating System

OSX, Linux

### What happened?

Using the `/shipments/.../smartrate/delivery_date` endpoint, UPS rates come back with a date format that is different from what easypost-go is expecting, resulting in a time parsing error.

Here is an example UPS rate returned from the above endpoint:
```
        {
            "easypost_time_in_transit_data": {
                "days_in_transit": {
                    "percentile_50": 1,
                    "percentile_75": 1,
                    "percentile_85": 1,
                    "percentile_90": 1,
                    "percentile_95": 2,
                    "percentile_97": 2,
                    "percentile_99": 2
                },
                "easypost_estimated_delivery_date": "2023-06-29",
                "planned_ship_date": "2023-06-28"
            },
            "rate": {
                "carrier": "UPS",
                "carrier_account_id": "ca_a4be3a46b7bc4657ab2f6c898e1d7f6e",
                "created_at": "2023-06-28T14:52:39Z",
                "currency": "USD",
                "delivery_date": "Thu, 29 Jun 2023 10:30:00 GMT",
                "delivery_date_guaranteed": true,
                "delivery_days": 1,
                "est_delivery_days": 1,
                "id": "rate_7f560e1598024218a94aad5eca72851e",
                "list_currency": "USD",
                "list_rate": 70.0,
                "mode": "production",
                "object": "Rate",
                "rate": 45.57,
                "retail_currency": "USD",
                "retail_rate": 71.31,
                "service": "NextDayAir",
                "shipment_id": "shp_087a68cd07074e9da840971ced82cca5",
                "updated_at": "2023-06-28T14:52:39Z"
            }
        },
```

The specific field that is causing the error is:
`"delivery_date": "Thu, 29 Jun 2023 10:30:00 GMT",`

Paths to fix:
1. Modify EasyPost's internal API to return a standard date format for all dates across carrier rates.
2. Modify easypost-go to handle these different date formats.

<img width="459" alt="image" src="https://github.com/EasyPost/easypost-go/assets/10538007/3288a199-0161-49f7-9cff-9a3ad852b4d2">

### What was expected?

The `GetShipmentEstimatedDeliveryDate` returns UPS rates without erroring.

coltmeister:
[...]

PR #184, opened 2023-06-29
[POC] Replace *time.Time usage with new DateTime class

nwithan8:
# Description

RE: #183 

- Handle any potential datetime format discrepancies coming from API by using new DateTime class instead of *time.Time
- Breaking change: Users will need to use `DateTime` instead of `time.Time` when setting times in request parameters (e.g. Pickup creation parameter set)

# Testing

- New unit test to confirm DateTime construction, de/serialization working as expected

# Pull Request Type


Justintime50 (commented):
question: what's the ultimate goal of this PR? To parse all sorts of different types of datetime strings without blowing up?

Do we want to require users to use a custom struct here instead of Go's builtin?

Do we want to open this up to all sorts of new types of strings when in reality we only wanted one type?

nwithan8:
> question: what's the ultimate goal of this PR? To parse all sorts of different types of datetime strings without blowing up?
> 
> Do we want to require users to use a custom struct here instead of Go's builtin?
> 
> Do we want to open this up to all sorts of new types of strings when in reality we only wanted one type?

1. Yes, exactly.
2. It's an unfortunate side-effect to ensure their code doesn't error out because of our API inconsistency (and one that could be undone if/when our API is no longer inconsistent). The new `DateTime` object inherits/extends the default `time.Time` struct, so there should be some backwards compatibility (as much as there really can be in Go, when inheritence isn't **really** a thing...). This was modeled after what the SaltBox team did in response to our prior SmartRate inconsistency: https://github.com/saltboxteam/easypost-go/commit/dce1d6fac5500fd046ff57d0855506ef023df07d#diff-9947c0020e50a9a546193173a4eb73b1b0e30eaadeac8c6aa5d8a28e78c14ec3
3. We can pare it back to just the two known types of date strings the API returns now. Was just trying to cover all our bases.

Justintime50:
I like the thoroughness, just want to make sure we've considered the ramifications. This only changes the deserializing of responses correct? What happens for users passing us dates? What I want to avoid is them now being able to pass us invalid dates because we loosened this up.

nwithan8:
> I like the thoroughness, just want to make sure we've considered the ramifications. This only changes the deserializing of responses correct? What happens for users passing us dates? What I want to avoid is them now being able to pass us invalid dates because we loosened this up.

In that case, specifically when creating a pickup here, since we use the same `Pickup` struct for both the request and response, the user would have to use our new `DateTime` type rather than `time.Time`. In the HTTP process, it would be serialized the same way the old `time.Time` class is serialized.

See: https://github.com/EasyPost/easypost-go/pull/184/files#diff-dcf9c624509270b270b9f2cb7706878bf577e089bc61dbebcfb412bfc49849e8L162

Justintime50:
As long as the final product is the same (when sending to the API), I'm fine with this. Again, want to ensure users can't send us new kinds of timestamps that the API won't know what to do with

nwithan8:
> As long as the final product is the same (when sending to the API), I'm fine with this. Again, want to ensure users can't send us new kinds of timestamps that the API won't know what to do with

via the `TestPickupCreate` cassette:

Previous recordings:
image


Recordings under the next system:
image


As you can see, the format of the date in the JSON payload sent to our API did not change. This PR does not contain a re-recorded cassette because it did not affect the payload.

Justintime50 (approved):
This seems fine, I'd like us to test as many of those cases as possible though, or if we know they'll never trigger, remove them

jchen293 (approved):

PR #186, opened 2023-07-07
[feat] Callbacks (hooks) audit pre-flight HTTP requests and post-flight HTTP responses

nwithan8:
# Description

- New optional `Hooks` property on a `Client` that contains pre-request and post-response event subscriptions
- New README example for using pre- and post-flight HTTP callbacks
- Add Google UUID package as dependency

# Testing

- New unit test to verify that pre- and post-flight hooks are called correctly.
- New unit test to verify that multiple pre- and multiple post-flight hooks can be called
- New unit test to verify that hooks can be subscribed and unsubscribed to.

# Pull Request Type


hooks.go
```diff
@@ -0,0 +1,113 @@
...
+
+// HookEvent is the base type for all hook events
+type HookEvent struct {
+}
+
+// HookEventSubscriber is the base type for all hook event subscribers.
+// ID needs to be unique in order to find and remove the subscriber
+type HookEventSubscriber struct {
+	ID string // need unique ID for removal
+}
```
jchen293, line 19: Question: why do we need these two struct?

hooks.go
```diff
@@ -0,0 +1,113 @@
...
+	Callback            RequestHookEventSubscriberCallback
+}
+
+// Execute executes the RequestHookEventSubscriberCallback function of the RequestHookEventSubscriber
+func (s RequestHookEventSubscriber) Execute(ctx context.Context, event RequestHookEvent) {
+	err := s.Callback(ctx, event)
+	if err != nil {
+		return
+	}
+}
```
jchen293, line 47:
```suggestion
// Executes the RequestHookEventSubscriberCallback function of the RequestHookEventSubscriber
func (s RequestHookEventSubscriber) Execute(ctx context.Context, event RequestHookEvent) {
	err := s.Callback(ctx, event)
	if err != nil {
		return
	}
}
```

Can we spell out what `s` is in the parameter?

hooks.go
```diff
@@ -0,0 +1,113 @@
...
+
+// HookEventSubscriber is the base type for all hook event subscribers.
+// ID needs to be unique in order to find and remove the subscriber
+type HookEventSubscriber struct {
+	ID string // need unique ID for removal
+}
+
+// RequestHookEvent is the data passed to a RequestHookEventSubscriberCallback function
+type RequestHookEvent struct {
+	HookEvent        // implements HookEvent
```
jchen293, line 23: I am not sure why we need this

hooks.go
```diff
@@ -0,0 +1,113 @@
...
+
+// Execute executes the ResponseHookEventSubscriberCallback function of the ResponseHookEventSubscriber
+func (s ResponseHookEventSubscriber) Execute(ctx context.Context, event ResponseHookEvent) {
+	err := s.Callback(ctx, event)
+	if err != nil {
+		return
+	}
+}
+
+// Hooks is a collection of HookEventSubscriber instances for various hooks available in the client
```
jchen293, line 79:
```suggestion
// Hooks are collections of HookEventSubscriber instances for various hooks available in the client
```

hooks.go
```diff
@@ -0,0 +1,113 @@
...
+type Hooks struct {
+	RequestHookEventSubscriptions  []RequestHookEventSubscriber // these are directly accessible by the user, but should not be modified directly (use Add/Remove methods)
+	ResponseHookEventSubscriptions []ResponseHookEventSubscriber
+}
+
+// AddRequestEventSubscriber adds a RequestHookEventSubscriber to the Hooks instance to be executed when a RequestHookEvent is fired
+func (h *Hooks) AddRequestEventSubscriber(subscriber RequestHookEventSubscriber) {
```
jchen293, line 86:
```suggestion
func (hook *Hooks) AddRequestEventSubscriber(subscriber RequestHookEventSubscriber) {
```

Nit: spell out the `h` in the parameter, this applies to below functions too

jchen293 (changes requested):
Can you also update the changelog?

hooks.go
nwithan8: For shared inheritance of the `Request` and `Response` events/subscribers

hooks.go
nwithan8: This is the data object that will be passed to the callback function, like we did in Java and .NET

jchen293 (approved):

ralexmatthews (approved):
LGTM 👍

PR #188, opened 2023-07-25
[feat] Add specific error types for API/HTTP-related errors

nwithan8:
# Description

- Add class and sub-class structure for improved error types in this library
  - `Error` is a server-side object (details about what went wrong)
  - `LibraryError` is an exception thrown by this library, the base of all custom exceptions
    - `ApiError` is an exception thrown by this library when an API/HTTP call fails, the base of all custom API exceptions
    - `LocalError` is an exception thrown by this library when a local error occurs, the base of all custom local exceptions
- Add sub-`ApiError` types for all possible HTTP status codes, deserialize response JSON into appropriate class instance
- Add sub-`LocalError` types for some local errors (not implemented except for webhook validation; coming in future PR)

# Testing

- Add unit tests to confirm API error determination and deserialization works
- Update some existing unit tests to except new specific API and local error types

# Pull Request Type


jchen293 (commented):
Need a changelog. We should probably make a branch for `v2` from master and then any breaking changes will be merged to the `master` branch

nwithan8:
> Need a changelog. We should probably make a branch for `v2` from master and then any breaking changes will be merged to the `master` branch

I'll build the CHANGELOG and UPGRADE GUIDE in a future PR.

error.go
```diff
@@ -72,22 +65,242 @@ func collectMessages(data interface{}, messages []string) []string {
...
+}
+
+type JsonNoDataError struct {
+	JsonError
+}
+
+type MissingParameterError struct {
+	LocalError
+}
+
```
Justintime50, line 133: thought: not a fan of the cascading errors of doom here. The multiple nested concept is difficult to follow.

error.go
```diff
@@ -72,22 +65,242 @@ func collectMessages(data interface{}, messages []string) []string {
...
 }
 
-var EndOfPaginationError = errors.New("there are no more pages to retrieve")
+type UnknownHttpError struct {
+	APIError
+}
+
+// FromErrorResponse returns an APIError-based object based on the HTTP response.
+// Do not pass a non-error response to this function.
+func FromErrorResponse(response *http.Response) error {
```
Justintime50, line 244: issue: can we rename this function? By the name alone I wouldn't know what this does. Verb - Action. Maybe something like `BuildErrorFromResponse`?

error.go
```diff
@@ -72,22 +65,242 @@ func collectMessages(data interface{}, messages []string) []string {
...
+	// build the base APIError object from the response
+	apiError := &APIError{
+		StatusCode: response.StatusCode,
+	}
+
+	// deserialize the response body into a temporary object
+	buf, _ := ioutil.ReadAll(response.Body)
+	aux := &struct {
```
Justintime50, line 252: nit: aux? Prefer `tmp` maybe or even `tmpError`?

I actually have no idea what this is supposed to mean? Can we be more explicit here?

error.go
```diff
@@ -72,22 +65,242 @@ func collectMessages(data interface{}, messages []string) []string {
...
+	switch response.StatusCode {
+	case 0:
+		return &ConnectionError{APIError: *apiError}
+	case 100, 101, 102, 103:
+		return &RetryError{APIError: *apiError}
+	case 300, 301, 302, 303, 304, 305, 306, 307, 308:
+		return &RedirectError{APIError: *apiError}
+	case 400:
+		return &BadRequestError{APIError: *apiError}
+	case 401, 403:
```
Justintime50, line 281: issue: 401 is unauthorized, 403 is forbidden and the distinction is important. Can we break these two up like we've done elsewhere?

error.go
```diff
@@ -72,22 +65,242 @@ func collectMessages(data interface{}, messages []string) []string {
...
+		apiError.Errors = []*Error{}
+	}
+
+	// return the appropriate error type based on the status code
+	switch response.StatusCode {
+	case 0:
+		return &ConnectionError{APIError: *apiError}
+	case 100, 101, 102, 103:
+		return &RetryError{APIError: *apiError}
+	case 300, 301, 302, 303, 304, 305, 306, 307, 308:
```
Justintime50, line 277: thought: can we (before the switch case) return RedirectError if it's 3xx at all?

Justintime50 (changes requested):
praise: nice work! Just a couple variable naming items and splitting up one error type that was missed and this should be golden.

error.go
nwithan8: Why not just keep it in the switch case?

error.go
Justintime50: right, the comment is the only reason I kind of new what this was; however, that comment doesn't follow the variable name throughout the rest of the code which is why having appropriately named vars really helps the further your travel vertically from the origin.

error.go
nwithan8: Resolved

error.go
nwithan8: Resolved

error.go
nwithan8: No, because if we come across an HTTP status code we aren't anticipating, we want that to be a different error.

error.go
nwithan8: Resolved

error.go
nwithan8: Resolved

Justintime50 (approved):

error.go
```diff
@@ -72,22 +65,271 @@ func collectMessages(data interface{}, messages []string) []string {
...
+		return &TimeoutError{APIError: *apiError}
+	case 422:
+		return &InvalidRequestError{APIError: *apiError}
+	case 429:
+		return &RateLimitError{APIError: *apiError}
+	case 500:
+		return &InternalServerError{APIError: *apiError}
+	case 503:
+		return &ServiceUnavailableError{APIError: *apiError}
+	case 504:
```
jchen293, line 330:
```suggestion
	case 502, 504:
```
Our API returns 502, let's add it so this will not fall to `UnknownHttpError`

jchen293 (commented):
Will there be a separate PR to use the `LibraryError`? Example, we should replace the error in those utils functions: https://github.com/EasyPost/easypost-go/blob/master/util.go

nwithan8:
> https://github.com/EasyPost/easypost-go/blob/master/util.go

Yes, a separate PR will implement all the local errors.

jchen293 (approved):

PR #192, opened 2023-07-27
[chore] Implement local error types and message constants

nwithan8:
# Description

- New `LocalError`-based error classes have been implemented and used throughout the code base
- Messages used inside the errors have been made as constants that end-users could compare against:
```go
assert.True(NoRatesFoundMatchingFilters, err.Error())
```

# Testing

# Pull Request Type


Justintime50 (commented):
Why is everything prepended with `New`? That may not be the interface we're after. If it's a naming issue, can we rename something else?

Or, is that intended for internal use only? If so, we should lowercase the `n` of New so it's not publicly accessible

nwithan8:
> Why is everything prepended with `New`? That may not be the interface we're after. If it's a naming issue, can we rename something else?
> 
> Or, is that intended for internal use only? If so, we should lowercase the `n` of New so it's not publicly accessible

It is for internal use only, and it is due to a naming thing. Good idea on the lowercase to make them private.

constants.go
```diff
@@ -9,3 +9,13 @@ var ApiDidNotReturnErrorDetails = "API did not return error details"
 var NoPagesLeftToRetrieve = "There are no more pages to retrieve"
 var MismatchWebhookSignature = "Webhook received did not originate from EasyPost or had a webhook secret mismatch"
 var MissingWebhookSignature = "Webhook does not contain a valid HMAC signature."
+var NoPaymentMethods = "No payment methods are set up. Please add a payment method and try again."
```
Justintime50, line 12: nit: can we sort this whole list?

referral_customer.go
```diff
@@ -141,12 +140,17 @@ func (c *Client) AddReferralCustomerCreditCard(referralCustomerApiKey string, cr
 func (c *Client) AddReferralCustomerCreditCardWithContext(ctx context.Context, referralCustomerApiKey string, creditCardOptions *CreditCardOptions, priority PaymentMethodPriority) (out *PaymentMethodObject, err error) {
 	stripeApiKeyResponse, err := c.retrieveEasypostStripeApiKey(ctx)
 	if err != nil || stripeApiKeyResponse == nil || stripeApiKeyResponse.PublicKey == "" {
-		return nil, errors.New("could not create Stripe token, please try again later")
+		return nil, &InternalServerError{
+			APIError: APIError{
+				Code:       "Could not create Stripe token, please try again later",
+				StatusCode: 0,
```
Justintime50, line 146: question: why a 0 status code here? I don't think that's what we want to be returning to users. Do we have custom logic that transforms that to something else? Why not return an appropriate 5xx error?

constants.go
nwithan8: Resolved

referral_customer.go
nwithan8: Because it's not really the response triggered by the HTTP failure, we're manufacturing this. The Status Code isn't shown anywhere, so it's mostly a placeholder. But we can make it 500 just to be consistent with what a InternalServerError would be.

Justintime50 (approved):

jchen293 (approved):

PR #208, opened 2024-01-08
[chore] Fix address verification unit test

nwithan8:
# Description

See https://github.com/EasyPost/easypost-csharp/pull/540 for more details

- Update examples submodule
- Improve `TestAddressCreateAndVerify` to better test target functionality
  - Unlike other libraries that will return the Address even if it is invalid from verification, due to how Go error checking works, the Address object is not returned, only the error.

# Testing

- Update unit test, re-record cassettes as needed

# Pull Request Type

### DIFF
--- a/address.go
+++ b/address.go
@@ -65,6 +65,7 @@ type Address struct {
 // CreateAddressOptions is used to specify verification options for address
 // creation.
 type CreateAddressOptions struct {
+	// TODO: These should be booleans, not strings, like the other libs
 	Verify       []string `json:"verify,omitempty"`
 	VerifyStrict []string `json:"verify_strict,omitempty"`
 }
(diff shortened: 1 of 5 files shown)


Justintime50 (approved):

address.go
```diff
@@ -65,6 +65,7 @@ type Address struct {
 // CreateAddressOptions is used to specify verification options for address
 // creation.
 type CreateAddressOptions struct {
+	// TODO: These should be booleans, not strings, like the other libs
```
jchen293, line 68: Could we include this fix as part of the next release?

jchen293 (approved):

address.go
nwithan8: This would be a breaking change, we should save this for a major release

PR #213, opened 2024-04-02
Add RefundInsurance function in insurance service

jchen293:
# Description

Add `RefundInsurance` function in the insurance service
# Testing

Added a new unit test

# Pull Request Type

### DIFF
--- a/tests/insurance_test.go
+++ b/tests/insurance_test.go
@@ -104,5 +104,5 @@ func (c *ClientTests) TestInsuranceRefund() {
 
 	assert.Equal(reflect.TypeOf(&easypost.Insurance{}), reflect.TypeOf(refundInsurance))
 	assert.True(strings.HasPrefix(refundInsurance.ID, "ins_"))
-	assert.Equal("100.00000", refundInsurance.Amount)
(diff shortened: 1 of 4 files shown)


tests/insurance_test.go
```diff
@@ -90,3 +90,19 @@ func (c *ClientTests) TestInsuranceGetNextPage() {
 		return
 	}
 }
+
+func (c *ClientTests) TestInsuranceRefund() {
+	client := c.TestClient()
+	assert, require := c.Assert(), c.Require()
+
+	insuranceData := c.fixture.BasicInsurance()
+	insuranceData.TrackingCode = "EZ1000000001"
```
Justintime50, line 99: question: why are we setting `TrackingCode` here? Is there not one in the fixture? If there is, we should remove the redundant line here

Justintime50 (approved):

tests/insurance_test.go
jchen293: In test mode, only the `EZ1000000001` tracking status is `pre-transit` which makes the insurance cancellable, the tracking code from the fixture is using a random tracking code.

nwithan8 (approved):

PR #224, opened 2024-07-10
feat: Account for new UPS endpoint in carrier account create/update functions

nwithan8:
# Description

- Adds new `CreateUpsCarrierAccount` and `UpdateUpsCarrierAccount` methods and related parameter structs
- Add new `InvalidFunctionError`, raised if users attempt to mix UPS and non-UPS function calls and data

This is the least destructive path to integrating the completely-different payload schema for UPS accounts. Outside of the new error that will be raised if users attempt to mix UPS and non-UPS data, there are no breaking changes (compilation) introduced with this PR.

# Testing

- New/updated unit tests for UPS and non-UPS creation/update functions
  - Updated cassettes as needed

# Pull Request Type

### DIFF
--- a/carrier.go
+++ b/carrier.go
@@ -11,8 +11,7 @@ type CarrierField struct {
 	Value      string `json:"value,omitempty"`
 }
 
-// CarrierFields contains the data for carrier account fields for production
-// and/or test credentials.
+// CarrierFields contains the data for carrier account fields for production and/or test credentials.
 type CarrierFields struct {
 	Credentials     map[string]*CarrierField `json:"credentials,omitempty"`
 	TestCredentials map[string]*CarrierField `json:"test_credentials,omitempty"`
@@ -20,8 +19,8 @@ type CarrierFields struct {
 	CustomWorkflow  bool                     `json:"custom_workflow,omitempty"`
 }
 
-// CarrierAccount encapsulates credentials and other information related to a
-// carrier account.
+// CarrierAccount encapsulates credentials and other information related to a carrier account.
+// This struct is also used as a parameter set for creating and updating non-UPS carrier accounts.
 type CarrierAccount struct {
 	ID               string                 `json:"id,omitempty"`
 	Object           string                 `json:"object,omitempty"`
@@ -39,8 +38,7 @@ type CarrierAccount struct {
 	BillingType      string                 `json:"billing_type,omitempty"`
 }
 
-// CarrierType contains information on a supported carrier. It can be used to
-// determine the valid fields for a carrier account.
+// CarrierType contains information on a supported carrier. It can be used to determine the valid fields for a carrier account.
 type CarrierType struct {
 	Object   string         `json:"object,omitempty"`
 	Type     string         `json:"type,omitempty"`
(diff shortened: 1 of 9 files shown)


carrier.go
```diff
@@ -50,34 +48,59 @@ type CarrierType struct {
...
+	AccountNumber string `json:"account_number,omitempty"`
+}
+
+type upsCarrierAccountCreationRequest struct {
+	Data *UpsCarrierAccountCreationParameters `json:"ups_oauth_registrations,omitempty"`
+}
+
+// UpsCarrierAccountUpdateParameters contains the parameters needed to update a UPS carrier account.
+type UpsCarrierAccountUpdateParameters struct {
+	// TODO: Add Description and Reference if/when API allows updating them.
```
Justintime50, line 68: suggestion: let's remove this TODO as it's uncertain if/when this will happen and instead create an internal ticket to address this holistically across all the libs. Then we can carry on the convo with the team that is building the API to determine if/when it can go down.

carrier.go
```diff
@@ -50,34 +48,59 @@ type CarrierType struct {
...
+}
+
+type upsCarrierAccountUpdateRequest struct {
+	Data *UpsCarrierAccountUpdateParameters `json:"ups_oauth_registrations,omitempty"`
+}
+
+func (c *Client) selectCarrierAccountCreationEndpoint(typ string) string {
 	for _, carrier := range getCarrierAccountTypesWithCustomWorkflows() {
-		if account.Type == carrier {
+		if typ == carrier {
```
Justintime50, line 78: question: is `type` a builtin for Go?

carrier.go
```diff
@@ -92,29 +115,60 @@ func (c *Client) GetCarrierTypesWithContext(ctx context.Context) (out []*Carrier
...
+	return c.CreateUpsCarrierAccountWithContext(context.Background(), in)
+}
+
+// CreateUpsCarrierAccountWithContext performs the same operation as CreateUpsCarrierAccount, but allows specifying a context that can interrupt the request.
+// Users cannot create non-UPS accounts with this function, must use CreateCarrierAccount. An error will be returned if the user tries to create a non-UPS account with this function.
+func (c *Client) CreateUpsCarrierAccountWithContext(ctx context.Context, in *UpsCarrierAccountCreationParameters) (out *CarrierAccount, err error) {
+	// Users cannot create non-UPS accounts with this function, must use CreateCarrierAccount
+	isUpsAccount := false
+	for _, carrier := range getUpsCarrierAccountTypes() {
+		if in.Type == carrier {
```
Justintime50, line 152: question: is there no way in Go to just check that a value exists in a range? Surely there is...

carrier.go
nwithan8: Yes, it will complain if you use it. Just like "klass" in Java, lol

image

carrier.go
nwithan8: There's slices, but that's only Go 1.21+: https://pkg.go.dev/slices#Contains
 
 Sometimes the good ol' loop is the best option.

carrier.go
```diff
@@ -50,34 +48,58 @@ type CarrierType struct {
...
+}
+
+type upsCarrierAccountCreationRequest struct {
+	Data *UpsCarrierAccountCreationParameters `json:"ups_oauth_registrations,omitempty"`
+}
+
+// UpsCarrierAccountUpdateParameters contains the parameters needed to update a UPS carrier account.
+type UpsCarrierAccountUpdateParameters struct {
+	AccountNumber string `json:"account_number,omitempty"`
+}
```
jchen293, line 69: Nit: add TODO comments here to allow `description` and `reference` when the endpoint allows it

carrier.go
```diff
@@ -92,29 +114,60 @@ func (c *Client) GetCarrierTypesWithContext(ctx context.Context) (out []*Carrier
...
+	isUpsAccount := false
+	for _, carrier := range getUpsCarrierAccountTypes() {
+		if in.Type == carrier {
+			isUpsAccount = true
+			break
+		}
+	}
+	if !isUpsAccount {
+		return nil, newInvalidFunctionError("users must use CreateCarrierAccount to create non-UPS accounts")
+	}
```
jchen293, line 158: Nit: move this into a private helper function to keep the function clean

jchen293 (approved):

carrier.go
Justintime50: I'd like us to avoid polluting external code with internal concerns. We need to address this internally with the API team to ensure it's even ever going to happen instead of alluding to it to users before we know.

PR #229, opened 2024-08-05
[chore] Use query for GET/DELETE requests, per HTTP spec

nwithan8:
# Description

- Update client logic to use query parameters for GET/DELETE requests and body for POST/PUT/PATCH requests
- Consolidate `.get`, `.post`, `.delete`, etc. functions to single `.do`
- Denote [URL and JSON key names](https://stackoverflow.com/a/19885276/13343799) for parameter serialization

No changes to end-user experience or user-facing function signatures

# Testing

- Re-record all cassettes (due to parameter schema changes, fixture updates, expiration)

# Pull Request Type


Justintime50 (changes requested):
Only blocking until I can take a solid look at this. You weren't kidding, this is beefy. Probably necessitates a breaking-change release, thoughts? We can do that shortly as we have a few other items that can pair with it if we decide to go that route.

client.go
```diff
@@ -97,26 +95,17 @@ func (c *Client) client() *http.Client {
 	return client
 }
 
-func (c *Client) convertOptsToURLValues(v interface{}) url.Values {
-	values, _ := query.Values(v)
-	return values
-}
```
jchen293, line 103: Hmm, so this function never works and we never convert the params into query URL for GET/DELETE requests?

jchen293 (approved):
lgtm, didn't find any typo or issue. Would be good to have another eyes on this PR

client.go
jchen293: I was referring to the `convertOptsToURLValues` function, which seems to have never worked before this PR? We use `convertOptsToURLValues` in the GET request body it was still sending the request data in the JSON body?

jchen293:
> Only blocking until I can take a solid look at this. You weren't kidding, this is beefy. Probably necessitates a breaking-change release, thoughts? We can do that shortly as we have a few other items that can pair with it if we decide to go that route.

I don't think it's a breaking change unless users have VCR cassettes with our Client Libs API interaction, then it could fail the tests. We already merged the Ruby PR for this change, so we have to make a call if we should hold off on the Ruby until the next breaking change.

client.go
nwithan8: It was used sporadically and inconsistently in upstream functions, which was likely what was causing some GET requests to be using body. Its new location ensures it will be applied to all GET requests.

Justintime50 (approved):
praise: Nice work here!

I would like to see the function suggestion investigated, I think it'd be worthwhile to make a single unified interface since there are hundreds of callsites that could benefit from it.

insurance.go
```diff
@@ -9,33 +9,33 @@ import (
...
-	ProviderID   string    `json:"provider_id,omitempty"`
-	ShipmentID   string    `json:"shipment_id,omitempty"`
-	TrackingCode string    `json:"tracking_code,omitempty"`
-	Status       string    `json:"status,omitempty"`
-	Tracker      *Tracker  `json:"tracker,omitempty"`
-	ToAddress    *Address  `json:"to_address,omitempty"`
-	FromAddress  *Address  `json:"from_address,omitempty"`
-	Fee          *Fee      `json:"fee,omitempty"`
-	Messages     []string  `json:"messages,omitempty"`
+	ID           string    `json:"id,omitempty" url:"id,omitempty"`
```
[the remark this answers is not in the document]
nwithan8: It doesn't appear as if you can supply a function for a [struct field tag](https://www.dolthub.com/blog/2024-02-07-go-tags/), it needs to be a static string during compile time.

We could potentially [implement our own struct field tag](https://medium.com/pragmatic-programmers/parsing-struct-tags-a99c5db9accd) that would under-the-hood handle both URL and JSON encoding, but I believe it's best to simply list both the [existing defined tags](https://go.dev/wiki/Well-known-struct-tags) rather than reinvent the wheel.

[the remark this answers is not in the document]
Justintime50: Dang, too bad. Thanks for looking!

PR #235, opened 2024-08-12
[bug] Fix error type inheritance

nwithan8:
# Description

We had previously incorrectly implemented "inheritance" (which really doesn't exist in Go) among our error types, which was causing "casting" issues if you wanted to check if a specific error (e.g. `InvalidRequestError`) was a subtype of a parent error type (e.g. `ApiError`).

This has been rectified by implementing [the `Unwrap` method](https://go.dev/blog/go1.13-errors#the-unwrap-method) from the standard `errors` package, which specifically handles casting for sub-types of errors.

Users can now build casting if-else or switch-case blocks to analyze which specific type of error was returned by a given function. Ex:

```golang
        // Print different messages based on the error type
	if err != nil {
		// LibraryError is the base of all errors thrown in this library, should match
		var libraryError *easypost.LibraryError
		if errors.As(err, &libraryError) {
			fmt.Println("Library error")
			fmt.Println(libraryError.Message)
		}
	
		// APIError is the base of all errors thrown due to the API, and a subtype of LibraryError, should match
		var apiError *easypost.APIError
		if errors.As(err, &apiError) {
			fmt.Println("API error")
			fmt.Println(apiError.Message)
			fmt.Println(apiError.StatusCode)
			fmt.Println(apiError.Errors)
		}

		// ForbiddenError is an error due to a 403 status code, and a subtype of APIError, should match
		var forbiddenError *easypost.ForbiddenError
		if errors.As(err, &forbiddenError) {
			fmt.Println("Forbidden error")
			fmt.Println(forbiddenError.Message)
			fmt.Println(forbiddenError.StatusCode)
			fmt.Println(forbiddenError.Errors)
		}

		// Alternate way of checking type if you don't need the resultant casted variable
		if errors.As(err, new(*easypost.ForbiddenError)) {
			// Do something, no access to casted error type
		}
	}
```

# Testing

- Updated unit tests accordingly
- Added new unit test to verify casting/inheritance tree for error types

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # CHANGELOG
 
+## Next Release
+
+- Fix inheritance tree for error types, allowing end-user to properly cast and check for specific error types and sub-types
+  - Some static error variables with non-dynamic error messages (e.g. `easypost.EndOfPaginationError`) have been removed and replaced with proper structs. Users relying on these variables to access the underlying error message should now use the error message directly (e.g. `easypost.NoPagesLeftToRetrieve`)
+    - Affected error variables and migration to static error messages:
+      - `easypost.EndOfPaginationError` -> `easypost.NoPagesLeftToRetrieve`
+      - `easypost.MissingWebhookSignatureError` -> `easypost.MissingWebhookSignature`
+      - `easypost.MismatchWebhookSignatureError` -> `easypost.MismatchWebhookSignature`
+
 ## v4.5.1 (2024-08-09)
 
 - Update HTTP logic to use query for GET/DELETE requests and body for POST/PUT/PATCH requests
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,6 @@
 ## Next Release
 
 - Fix inheritance tree for error types, allowing end-user to properly cast and check for specific error types and sub-types
-  - Some static error variables with non-dynamic error messages (e.g. `easypost.EndOfPaginationError`) have been removed and replaced with proper structs. Users relying on these variables to access the underlying error message should now use the error message directly (e.g. `easypost.NoPagesLeftToRetrieve`)
-    - Affected error variables and migration to static error messages:
-      - `easypost.EndOfPaginationError` -> `easypost.NoPagesLeftToRetrieve`
-      - `easypost.MissingWebhookSignatureError` -> `easypost.MissingWebhookSignature`
-      - `easypost.MismatchWebhookSignatureError` -> `easypost.MismatchWebhookSignature`
 
 ## v4.5.1 (2024-08-09)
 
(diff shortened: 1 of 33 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # CHANGELOG
 
+## Next Release
+
+- Fix inheritance tree for error types, allowing end-user to properly cast and check for specific error types and sub-types
+  - Some static error variables with non-dynamic error messages (e.g. `easypost.EndOfPaginationError`) have been removed and replaced with proper structs. Users relying on these variables to access the underlying error message should now use the error message directly (e.g. `easypost.NoPagesLeftToRetrieve`)
+    - Affected error variables and migration to static error messages:
+      - `easypost.EndOfPaginationError` -> `easypost.NoPagesLeftToRetrieve`
+      - `easypost.MissingWebhookSignatureError` -> `easypost.MissingWebhookSignature`
+      - `easypost.MismatchWebhookSignatureError` -> `easypost.MismatchWebhookSignature`
```
Justintime50, line 10: question: can we not just swap in the structs and keep the names? At a minimum, we should probably deprecate and not remove if there are breaking changes to the user.

error.go
```diff
@@ -91,15 +91,37 @@ func (e *LibraryError) Error() string {
 
 // LocalError represents an error caused by the EasyPost library itself, such as validation or JSON serialization issues.
 type LocalError struct {
-	LibraryError
+	LibraryError // subtype of LibraryError
+}
+
+// Unwrap returns the underlying LibraryError object.
```
salfonso-easypost, line 97:
```suggestion
// Unwrap returns the underlying LibraryError error.
```

error.go
```diff
@@ -91,15 +91,37 @@ func (e *LibraryError) Error() string {
...
+
+// newEndOfPaginationError returns a new EndOfPaginationError object.
+func newEndOfPaginationError() *EndOfPaginationError {
+	return &EndOfPaginationError{LocalError{LibraryError{Message: NoPagesLeftToRetrieve}}}
+}
 
 // FilteringError is raised when there is an issue while running a filtering operation.
 type FilteringError struct {
-	LocalError
+	LocalError // subtype of LocalError
```
salfonso-easypost, line 119:
```suggestion
	*LocalError // subtype of LocalError
```

update all errors that have an error subtype to be a pointer to the error instead.

salfonso-easypost: not convinced this is entirely necessary

CHANGELOG.md
nwithan8: Discussing offline

error.go
nwithan8: Resolved

error.go
nwithan8: This seems to break things in my IDE, and I don't see any documentation about Go "inheritance" that suggests this.

Justintime50 (approved):

jchen293 (approved):

PR #241, opened 2025-02-21
feat: adding custom headers support for the webhook object.

codingminions:
# Description

Add support for users of this client-library to input custom headers attribute with the webhook object. We allow up to three custom headers, each with a key and value under 1000 ASCII characters. The header keys and values are stored and returned as plaintext.

The custom_headers attribute in the webhook object is a list of key-value pairs, where each entry consists of: 

Example Usage of the custom headers with this client-lib:
```
   header1 := easypost.WebhookCustomHeader{Name: "X-Custom", Value: "Header"}

    webhook, err := client.CreateWebhookWithDetails(
        &easypost.CreateUpdateWebhookOptions{
            URL: "https://example.com/webhook",
            CustomHeaders: []easypost.WebhookCustomHeader{header1},
    }); 
    
    if err != nil {
        log.Fatalf("Error creating webhook: %v", err.Error())
    } else {
        for _, header := range webhook.CustomHeaders {
            fmt.Printf("Header: %s\n", header.Name)
        }
    }
  ```
# Testing

 - This PR will most likely fail the CI build but we are aware of why at the moment. 
 - New unit tests have been added to test Create/Update of custom headers in conjunction with existing attributes of the webhook (check `tests/webhook_test.go` on this PR)
 - Re-recorded all existing Webhook test cassettes given the change in the client fixtures data along with recording VCR for new tests. 
 - The expiration time period for cassette has been changed to `365` from `180`
 

# Pull Request Type

# Additional Notes 

 - There are two new tests(`webhook_test.go: TestWebhookUpdateWithCustomHeaders, TestWebhookUpdateWithCustomHeadersAndSecret`  which are failing during update of custom headers:
 
     _Why_: The current implementation of the update method for webhooks might be creating an erroneous payload for the monolith failing our assertion that the updated webhooks contain the updated params. 
     
     _Action Item_:  A phab ticket has been raised(`T86916`) which will be looked by the SDK team. In case, a decision is made to keep the existing behavior of update, the new test cases will be revisited for success.
     
- @Justintime50 one of your pointers was: "Add the secret to the create/update tests as we weren’t previously sending those but should have been"
   
   I think there are tests in the `webhook_test.go` that already covers scenarios of creating and updating webhook with secrets?

### CODE REVIEW

File: tests/fixture_test.go [hunk 1]
```diff
@@ -210,15 +208,50 @@ func (fixture *Fixture) EventBody() []byte {
...
-	return readFixtureData().WebhookSecretString
+	return readFixtureData().Webhooks["secret"].(string)
 }
 
 func (fixture *Fixture) WebhookUrl() string {
-	return readFixtureData().WebhookURL
+	return readFixtureData().Webhooks["url"].(string)
+}
+
+func (fixture *Fixture) WebhookCustomHeaders() ([]easypost.WebhookCustomHeader, error) {
```
Justintime50, line 222: thought: I don't believe we need all this logic and especially not the error handling. These fixtures only intended purpose are to load data from the examples submodule files. We'd expect them to be properly formatted when accessed and therefore shouldn't need all the logic here.

tests/webhook_test.go
```diff
@@ -182,11 +182,123 @@ func (c *ClientTests) TestValidateWebhookMissingSecret() {
...
+	client := c.TestClient()
+	assert, require := c.Assert(), c.Require()
+
+	customHeaders, err := c.fixture.WebhookCustomHeaders()
+	require.NoError(err)
+
+	webhook, err := client.CreateWebhookWithDetails(
+		&easypost.CreateUpdateWebhookOptions{
+			URL:           c.fixture.WebhookUrl(),
+			CustomHeaders: customHeaders,
```
Justintime50, line 204: suggestion: I'd prefer that the `CustomHeaders` get passed in the `CreateWebhookWithDetails` test instead of needing a separate test here (see other client libraries). We are ultimately looking to keep the test suite as simple as possible and our create tests usually attempt to pass in happy-path params as a single test. Same for the update test.

So ultimately we'd want no new tests added, the secret and custom header params to be added to the existing create and update tests, and asserting that the custom header name/value is in the response (see other PRs)

Justintime50 (changes requested):
Nice work here! A couple small suggestions I'd like to see investigated to reduce the size/scope of this PR (we looove simple for the client libs).

As to the failing tests and the update function - yes. Please unwrap the parameters when sending the payload through. That code was written in error (following convention) but since to this point there weren't really any params to pass it wasn't caught. We'll release that bug fix alongside the custom header feature (note: some of the other lang PRs had to do the same so you're good to make that change here)

codingminions:
Hey @Justintime50 I have addressed the feedback comments and added the bug fix in the new commit. Please take a look and let me know if you have any questions for me. Thanks for the review!

Justintime50 (approved):
Great work here!

